PR #4, opened 2019-05-10
Leader election enabled.

georgekuruvillak:
This PR adds leader election to dependency-watchdog so that multiple instance will not adversely affect the correct behaviour.
Fixes #

**Release note**:

```improvement operator

```


main.go
```diff
@@ -86,24 +94,67 @@ func main() {
...
+	ctx := context.TODO()
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:          rl,
+		LeaseDuration: controller.LeaderElection.LeaseDuration.Duration,
+		RenewDeadline: controller.LeaderElection.RenewDeadline.Duration,
+		RetryPeriod:   controller.LeaderElection.RetryPeriod.Duration,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: run,
+			OnStoppedLeading: func() {
+				klog.Fatalf("leaderelection lost")
```
amshuman-kr: Shouldn't we stop the controller?

PR #7, opened 2019-11-21
Introduced cobra commands.

amshuman-kr:
1. The root command works exactly like before.
I.e. it watches the endpoint objects and kicks the
dependant pods in `CrashloopBackoff`.
2. The probe sub-command probes kube-apiservers
using internal and external IP kubeconfigs and
scales the dependant scale subresources up and down.

Both the root command and the probe sub-command support
managing a single namespace as well as all namespaces.

Partially Fixes [#1447@gardener](https://github.com/gardener/gardener/issues/1447). Needs to be integrated via [#1641@gardener](https://github.com/gardener/gardener/pull/1641).

I have moved the repo from dep to go modules. Also, I have removed any references to the `k8s.io/kubernetes` package which is not [recommended](https://github.com/golang/go/issues/32776#issuecomment-505607726).

**Release note**:

```noteworthy user
Introduced cobra commands.
1. The root command works exactly like before.
I.e. it watches the endpoint objects and kicks the
dependant pods in CrashloopBackoff.
2. The probe sub-command probes kube-apiservers
using internal and external IP kubeconfigs and
scales the dependant scale subresources up and down.

Both the root command and the probe sub-command support
managing a single namespace as well as all namespaces.
```


amshuman-kr:
cc @georgekuruvillak @rfranzke

pkg/scaler/prober.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (p *prober) isHealthy(pr *probeResult) bool {
+	return pr.lastError == nil && pr.resultRun >= p.successThreshold
+}
+
+func (p *prober) isUnhealthy(pr *probeResult) bool {
+	return pr.lastError != nil && pr.resultRun >= p.failureThreshold
+}
+
+func (p *prober) probe() error {
```
georgekuruvillak: There is an edge case where internal probe may not be both unhealthy and healthy. In that case, we shouldn't do the external probe as it can lead to scale up.

georgekuruvillak: Comment on the implementation would be good as well

pkg/restarter/restarter.go
```diff
@@ -201,7 +187,7 @@ func (c *Controller) processEndpoint(key string) error {
 		utilruntime.HandleError(fmt.Errorf("invalid resource key: %s", key))
 		return nil
 	}
-	if namespace != c.serviceDependants.Namespace {
+	if c.serviceDependants.Namespace != "" && namespace != c.serviceDependants.Namespace {
```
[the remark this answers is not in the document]
amshuman-kr: Not really. It is just an optimisation where enqueuing can be skipped if possible.

pkg/scaler/prober.go
amshuman-kr: 👍 Changed the logic handle the edge case.

amshuman-kr: Added comments to document the logic.

amshuman-kr:
@georgekuruvillak Thanks for the review comments. I have tried to address all of them. Can you PTAL?

PR #16, opened 2020-05-28
Fix/minimise prober restarts

### LINKED ISSUE #15: [BUG] Probe: Reduce load on the host apiserver during namespace/secret update
**Describe the bug:**
The dependency watchdog probe watches/reconciles the namespaces and secrets to re-configure/setup probes for the individual shoot namespaces to be probed.

Such reconciliation is not optimal right now as it treats each update to the namespace/secret as potentially a change to the kubeconfig secrets and proceeds to explicitly retrieve the secrets from the host apiserver and re-configure the probes.

This is suboptimal and can cause high load on the host apiserver and might end up throttling (see the log below) the probe process itself leading to many of the namespaces' shoot apiserver not probed for substantial period of time (> `2m`) which could be enough time for the individual `kube-controller-manager`s to act (potentially erroneously due to apiserver loadbalancer issues) on unresponsive nodes and mark them as `NotReady`.

**Expected behaviour:**
Individual probes for the different namespaces' shoot apiservers should be up as long as possible and should be reconfigured only if the kubeconfig secrets actually changed (with the content of the kubeconfig changing).

The down time for the individual probes should be as little as possible with secrets being loaded from the local informer cache rather than using a get call on the host apiserver (thus also reducing the load on the host apiserver).

**Logs:**
```
Throttling request took 196.193411ms, request: GET:https://...:443/api/v1/namespaces/.../secrets/dependency-watchdog-external-probe
```

amshuman-kr:
Probes are now restarted only if any of the target kubeconfigs actually changed. Also, the kubeconfig secrets are loaded from the informer rather than hitting the host apiserver (seed) with GET calls. Also, logged OS signal for debugging.

Fixes #15 

I had to revendor to the latest client-go and k8s api to resolve a nasty vendoring problem which also needed ugrading to golang 1.13 as in #17.

**Release note**:

```improvement operator
Probes are now restarted only if any of the target kubeconfigs actually changed. Also, the kubeconfig secrets are loaded from the informer rather than hitting the host apiserver (seed) with GET calls. Also, logged OS signal for debugging.
```


pkg/restarter/restarter.go
```diff
@@ -191,7 +191,7 @@ func (c *Controller) processEndpoint(key string) error {
 		return nil
 	}
 
-	ep, err := c.clientset.CoreV1().Endpoints(namespace).Get(name, metav1.GetOptions{})
+	ep, err := c.clientset.CoreV1().Endpoints(namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
georgekuruvillak, line 194: Making a context in `processWorkItem` and passing to `processEndpoint` helps reuse the context.

pkg/restarter/restarter.go
```diff
@@ -266,7 +266,7 @@ func (c *Controller) shootDependentPodsIfNecessary(ctx context.Context, namespac
 
 	for {
 		retry, err := func() (bool, error) {
-			w, err := c.clientset.CoreV1().Pods(namespace).Watch(metav1.ListOptions{
+			w, err := c.clientset.CoreV1().Pods(namespace).Watch(context.TODO(), metav1.ListOptions{
```
georgekuruvillak, line 269: Same as above.

pkg/restarter/restarter.go
```diff
@@ -312,13 +312,13 @@ func (c *Controller) shootDependentPodsIfNecessary(ctx context.Context, namespac
 
 func (c *Controller) processPod(pod *v1.Pod) error {
 	// Validate pod status again before shoot it out.
-	po, err := c.clientset.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
+	po, err := c.clientset.CoreV1().Pods(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
```
georgekuruvillak, line 315: Same as above.

pkg/restarter/restarter.go
```diff
@@ -312,13 +312,13 @@ func (c *Controller) shootDependentPodsIfNecessary(ctx context.Context, namespac
...
 	if err != nil {
 		return fmt.Errorf("error getting pod %s", pod.Name)
 	}
 	if !ShouldDeletePod(po) {
 		return nil
 	}
 	klog.Infof("Deleting pod: %v", po.Name)
-	return c.clientset.CoreV1().Pods(po.Namespace).Delete(po.Name, &metav1.DeleteOptions{})
+	return c.clientset.CoreV1().Pods(po.Namespace).Delete(context.TODO(), po.Name, metav1.DeleteOptions{})
```
georgekuruvillak, line 323: Same as above.

pkg/scaler/prober.go
```diff
@@ -68,50 +74,64 @@ type probeResult struct {
 	resultRun int32
 }
 
-func (p *prober) getClientFromSecret(secretName string) (kubernetes.Interface, error) {
+// getClientFromSecret constructs a Kubernetes client based on the supplied secret and
+// return it along with the SHA256 checksum of the kubeconfig in the secret
+// but only if the SHA256 checksum of the kubeconfig in the secret differs from oldSHA.
+func (p *prober) getClientFromSecret(secretName string, oldSHA []byte) (kubernetes.Interface, []byte, error) {
```
georgekuruvillak, line 80: When is this created. Why do we need this logic to avoid creating a k8s client if the secret didn't change?

pkg/restarter/restarter.go
amshuman-kr: After the escape from vendor hell, my brain was too frayed to do this 😅 
:+1: Will make the change.

amshuman-kr:
@georgekuruvillak I have addressed the context re-use. Can you PTAL?

gardener-botmanager:
@georgekuruvillak Command /lgtm does not have any arguments.

PR #23, opened 2020-06-29
Run probe only for clusters that are in ready state

ggaurav10:
Probe is stopped for clusters that are hibernating, or in hibernation or are still waking up from hibernation

Fixes #

**Release note**:

```improvement operator
Probe is stopped for clusters that are hibernating, or in hibernation or are still waking up from hibernation
```


pkg/scaler/prober.go
```diff
@@ -156,8 +158,16 @@ func (p *prober) tryAndRun(prepareRun func() (stopCh <-chan struct{})) error {
 	internalClient, internalSHA, internalErr := p.getClientFromSecret(p.probeDeps.Probe.Internal.KubeconfigSecretName, p.internalSHA)
 	externalClient, externalSHA, externalErr := p.getClientFromSecret(p.probeDeps.Probe.External.KubeconfigSecretName, p.externalSHA)
 
-	// Run a fresh prober goroutine if any one of the secrets changed
-	if internalErr == nil || externalErr == nil {
+	if apierrors.IsNotFound(internalErr) || apierrors.IsNotFound(externalErr) {
```
amshuman-kr, line 161: I was thinking over the week-end if we should change it to the following to cancel the probe if any error other than `AlreadyExists` occurs.
```suggestion
	if !apierrors.IsAlreadyExists(internalErr) && !apierrors.IsAlreadyExists(externalErr) {
```

pkg/scaler/prober.go
```diff
@@ -156,8 +158,16 @@ func (p *prober) tryAndRun(prepareRun func() (stopCh <-chan struct{})) error {
...
+		cancelFn()
+
+		if internalErr != nil {
+			return internalErr
+		}
+
+		return externalErr
+	} else if internalErr == nil || externalErr == nil {
```
amshuman-kr, line 169: Then this could be a simple `else`.

```suggestion
	} else {
```

pkg/scaler/prober.go
```diff
@@ -260,6 +270,31 @@ func (p *prober) getProbeResultLabels(pr *probeResult) prometheus.Labels {
...
+		decoder, err := extensionscontroller.NewGardenDecoder()
+		if err != nil {
+			return nil
+		}
+
+		shoot, err := extensionscontroller.ShootFromCluster(decoder, cluster)
+		if err != nil {
+			return nil
+		}
+		if shoot.Spec.Hibernation == nil || shoot.Spec.Hibernation.Enabled == nil || (*shoot.Spec.Hibernation.Enabled == false && shoot.Status.IsHibernated == false) {
```
amshuman-kr, line 291: Shouldn't this be the following?

```suggestion
		if (shoot.Spec.Hibernation == nil || shoot.Spec.Hibernation.Enabled == nil || *shoot.Spec.Hibernation.Enabled == false) && shoot.Status.IsHibernated == false {
```

pkg/scaler/utils.go
```diff
@@ -57,3 +59,25 @@ func isRateLimited(err error) bool {
 	const prefix = "rate: "
 	return strings.HasPrefix(err.Error(), prefix)
 }
+
+func shootWokeUp(old, new *gardenerv1alpha1.Cluster) bool {
```
amshuman-kr, line 63: Shouldn't this be more `shootInHibernationTransition`?

pkg/scaler/utils.go
```diff
@@ -57,3 +59,25 @@ func isRateLimited(err error) bool {
...
+	oldShoot, err := extensionscontroller.ShootFromCluster(decoder, old)
+	if err != nil {
+		return false
+	}
+	newShoot, err := extensionscontroller.ShootFromCluster(decoder, new)
+	if err != nil {
+		return false
+	}
+
+	if oldShoot.Status.IsHibernated == true && newShoot.Status.IsHibernated == false {
```
amshuman-kr, line 78: Shouldn't we check the spec also here?

pkg/scaler/prober.go
amshuman-kr: Also, shouln't we have a `shootWokeUp` or `shootInHibernationTransition` check here? Please note that this code-path might be reached from the `secretsInformers` event as well.

ggaurav10: > if we should change it to the following to cancel the probe if any error other than AlreadyExists occurs

hmm...makes sense. So if there is a transient network error, would the probe start again for the namespace when the error goes away (even if there are no updates to secrets or cluster state)?

>shouln't we have a shootWokeUp or shootInHibernationTransition check here?

in `probe()` we have already added check to close a probe if shoot is not up and ready. would we still need such a check here to close a probe? wdyt?

pkg/scaler/prober.go
ggaurav10: then would it result in restart of the probe even if both the secrets "already exist"?

pkg/scaler/utils.go
ggaurav10: This function is only to check the type of `update` event on the `cluster`. Returns true only if the `status` changed from "sleeping/waking up" to "awake/ready". I will change the name of the function to better reflect its purpose.

pkg/scaler/utils.go
ggaurav10: This function is only to check the type of `update` event on the `cluster`. Returns true only if the `status` changed from "sleeping/waking up" to "awake/ready". If the cluster is in any other transition, then it would result false at which point, we wouldn't want to "enqueue" the cluster key. wdyt?

pkg/scaler/utils.go
```diff
@@ -57,3 +60,36 @@ func isRateLimited(err error) bool {
...
+	if err != nil {
+		klog.V(4).Infof("Error getting new shoot version from cluster: %v. Err: %v", new.Name, err)
+		return false
+	}
+
+	if oldShoot.Status.IsHibernated != newShoot.Status.IsHibernated {
+		return true
+	}
+
+	if (oldShoot.Spec.Hibernation == nil && newShoot.Spec.Hibernation != nil) ||
```
amshuman-kr, line 86: Can't we use `reflect.DeepEqual` here?

ggaurav10: just wanted to avoid updates to other fields in the `Hibernation` field.
It can look like this:
```yaml
hibernation:
    enabled: false
    schedules:
      - start: '00 20 * * 1,2,3,4,5,6,0'
        end: '30 08 * * 1,2,3,4,5,6,0'
        location: Europe/Berlin
      - start: '30 05 * * 1,2,3,4,5,6,0'
        end: '30 01 * * 1,2,3,4,5,6,0'
        location: Europe/Berlin
```
wdyt?

PR #37, opened 2022-01-05
Switch leader election to endpointleases

ary1992:
/kind enhancement

Switch default leader election resource lock from `endpoints` to `endpointsleases`

Part of #4742

**Release note**:

```improvement operator
Switch default leader election resource lock for `dependency-watchdog` from `endpoints` to `endpointsleases`
```

```
Components that deploy the `dependency-watchdog` will now have to adapt the RBAC rules to allow `dependency-watchdog` to maintain its leader election resource lock in `leases` as well.
```

### DIFF
--- a/pkg/restarter/restarter.go
+++ b/pkg/restarter/restarter.go
@@ -15,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
 
 	"github.com/gardener/dependency-watchdog/pkg/multicontext"
@@ -42,6 +43,9 @@ func NewController(clientset kubernetes.Interface,
 		serviceDependants: serviceDependants,
 		watchDuration:     watchDuration,
 		Multicontext:      multicontext.New(),
+		LeaderElection: componentbaseconfigv1alpha1.LeaderElectionConfiguration{
+			ResourceLock: resourcelock.EndpointsLeasesResourceLock,
+		},
 	}
 	componentbaseconfigv1alpha1.RecommendedDefaultLeaderElectionConfiguration(&c.LeaderElection)
 	c.endpointInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
--- a/pkg/scaler/scaler.go
+++ b/pkg/scaler/scaler.go
@@ -13,6 +13,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
 
 	"github.com/gardener/dependency-watchdog/pkg/multicontext"
@@ -54,6 +55,9 @@ func NewController(clientset kubernetes.Interface,
 		stopCh:                 stopCh,
 		probeDependantsList:    probeDependantsList,
 		Multicontext:           multicontext.New(),
+		LeaderElection: componentbaseconfigv1alpha1.LeaderElectionConfiguration{
+			ResourceLock: resourcelock.EndpointsLeasesResourceLock,
+		},
 	}
 	componentbaseconfigv1alpha1.RecommendedDefaultLeaderElectionConfiguration(&c.LeaderElection)
 	c.clusterInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{


ashwani2k (commented):
Thanks @ary1992  for the PR.
Looks fine, however I was wondering won't  this change would also require us to adapt how the resource locks are acquired currently in the code as done here for [root cmd](https://github.com/gardener/dependency-watchdog/blob/96f73d5dc311958f5cb6e914fbda10b8bbc095c9/cmd/root.go#L176-#L184 )  and [probe cmd](https://github.com/gardener/dependency-watchdog/blob/master/cmd/probe.go#L122-#L130).

Or this is not relevant?

ary1992:
> Thanks @ary1992 for the PR. Looks fine, however I was wondering won't this change would also require us to adapt how the resource locks are acquired currently in the code as done here for [root cmd](https://github.com/gardener/dependency-watchdog/blob/96f73d5dc311958f5cb6e914fbda10b8bbc095c9/cmd/root.go#L176-#L184) and [probe cmd](https://github.com/gardener/dependency-watchdog/blob/master/cmd/probe.go#L122-#L130).
> 
> Or this is not relevant?

`resourcelock.New()` in  [root cmd](https://github.com/gardener/dependency-watchdog/blob/96f73d5dc311958f5cb6e914fbda10b8bbc095c9/cmd/root.go#L176-#L184) and [probe cmd](https://github.com/gardener/dependency-watchdog/blob/master/cmd/probe.go#L122-#L130) , takes first parameter as `locktype` which we have already defined in Controller with [restarter](https://github.com/gardener/dependency-watchdog/blob/a39609e6ae73e296fdc60cc03b24dc26e83a6208/pkg/restarter/restarter.go#L47) and [scaler](https://github.com/gardener/dependency-watchdog/blob/a39609e6ae73e296fdc60cc03b24dc26e83a6208/pkg/scaler/scaler.go#L59). Hence the resource lock acquired at `probe cmd` and `root cmd` will be of type `endpointsleases` instead of `endpoints`.

ashwani2k:
Thanks for the explanation. I understand that, but I was more referring to the adaptation as per the parent issue [ Umbrella Issue 4272](https://github.com/gardener/gardener/issues/4742) it is mentioned -- 
>To preserve the backwards-compatibility, resource lock for migration purposes (endpointsleases, configmapsleases) should be used when switching from the legacy resource locks (endpoints, configmaps).

So shouldn't we be using `configmapsLeases` instead of `endpointsLeases` here?
Also I see similar change in done for configmaps here in  [#PR8107](https://github.com/kubernetes/ingress-nginx/pull/8107/files#diff-366a097c35e219655414908a9106c9b0be5f0847df71aae8617f2892855a9a63L96).

ashwani2k:
I believe we will need to adapt the ClusterRoles definitions defined for dependency-watchdog [here ](https://github.com/gardener/gardener/blob/889f5506e0be3342e38207b2615e54e719ff3d66/pkg/operation/botanist/component/dependencywatchdog/bootstrap.go#L333 )

ary1992:
> I believe we will need to adapt the ClusterRoles definitions defined for dependency-watchdog [here](https://github.com/gardener/gardener/blob/889f5506e0be3342e38207b2615e54e719ff3d66/pkg/operation/botanist/component/dependencywatchdog/bootstrap.go#L333)

While testing the code changes in `dependency-watchdog` we have added the rule for `leases` [here](https://github.com/gardener/gardener/blob/889f5506e0be3342e38207b2615e54e719ff3d66/pkg/operation/botanist/component/dependencywatchdog/bootstrap.go#L174) . Will the changes be required at both the places.

As the RBAC rules are part of `gardener/gardener` , first the current changes needs to be merged in `gardener/dependency-watchdog`. Then only we can make the corresponding changes to RBAC. Please correct me if I am wrong.

ashwani2k:
>While testing the code changes in dependency-watchdog we have added the rule for leases here . Will the changes be required at both the places.
No I think we just need to change the role. So it's fine. 

I couldn't check the link you have mentioned as its pointing to a line https://github.com/gardener/gardener/blob/889f5506e0be3342e38207b2615e54e719ff3d66/pkg/operation/botanist/component/dependencywatchdog/bootstrap.go#L174 
> As the RBAC rules are part of `gardener/gardener` , first the current changes needs to be merged in `gardener/dependency-watchdog`. Then only we can make the corresponding changes to RBAC. Please correct me if I am wrong.

Yes that's correct path. I was just raising it here for completeness of the story from dependency-watchdog point of view. 
All good then. 

Maybe @amshuman-kr  if you happen to verify it once. 

/lgtm

acumino:
@ialidzhikov `If the leader election resource lock is not configurable via flag, then let's also make it configurable` is this required here or with `dependency-watchdog` also we adapt the default resource lock used in the `dependency-watchdog` similar to MCM.

ialidzhikov:
For MCM the leader election resource was already configurable, hence we only had to adapt its default value.
If it is not configurable for dependency-watchdog, we could make it. 👍

ialidzhikov:
@ary1992 , let's leave the part with making the resource-lock configurable our of this PR.

ashwani2k (approved):
/lgtm

PR #41, opened 2022-03-07
Handled behaviour for deployments when not available in the cluster

### LINKED ISSUE #40: [BUG] Dependency watchdog probe should ignore the deployments in case they don't exist in the landscape
**Describe the bug:**
Currently with the release 0.7.0 it is possible to define a `scaleRefDependsOn` to provide a list of dependency. 
If we define a dependency to say a component `cluster-autoscaler` as a dependency to `machine-controller-manager`. 
Currently as per design - 
```
Every shoot cluster that has at least one worker pool with minimum < maximum nodes configuration will get a cluster-autoscaler deployment. 
```
However in the config definition defined with release v0.7.0 this lead to error logs with failed to get deployment in the logs which reflects it as a failed condition. 

**Expected behaviour:**
If `cluster-autoscaler` is not enabled for a shoot cluster due to above defined reasons, we expect DWD to just report skipping of a deployment in the logs instead of marking it as failure. 

Create a shoot with a worker configuration of min=max. This will not enable cluster-autoscaler in the shoot control plane. 

**Logs:**
You will see in the logs of DWD probe errors like 
```
0304 07:15:46.904892       1 prober.go:405] Scaling up dependents of shoot-kube-apiserver/shoot--foo--bar: apps/v1.Deployment/cluster-autoscaler: replicas=1: failed
E0304 07:15:46.906923       1 prober.go:452] Scaling up dependents of shoot-kube-apiserver/shoot--foo--bar: apps/v1.Deployment/cluster-autoscaler: error getting deployments.apps: deployments/scale.apps "cluster-autoscaler" not found
E0304 07:15:46.906938       1 prober.go:500] Scaling up dependents of shoot-kube-apiserver/shoot--foo--bar: apps/v1.Deployment/cluster-autoscaler: Could not get target reference: deployments/scale.apps "cluster-autoscaler" not found
E0304 07:15:46.906943       1 prober.go:501] Scaling up dependents of shoot-kube-apiserver/shoot--foo--bar: apps/v1.Deployment/cluster-autoscaler: replicas=1: failed
```

ashwani2k:
This PR handles skipping the scaling operation for deployment when they are not available in the cluster. 
The logs are also refined to avoid noise and clutter. 

Fixes #40 

1. The PR upgrades GO dependency to 1.17
2. There is also a delay introduced before checking dependents for desire availability. This is introduced as checking for availability immediately after scaling operation always failed as it takes some time for the deployment to become ready.
Currently it is a hacky way with 2 sec delay as per what was observed from the tests. 
Without this delay also the scaling operation happens but it is honored in the next reconciliation cycle. This creates an overall delay of 30s scaling delay per deployment requiring scaling.

**Release note**:

```bugfix operator
Dependency watchdog shall not ignore scaling operations on deployment although defined in the config if they are not configured/enabled in the cluster. 
```


pkg/scaler/prober.go
```diff
@@ -349,22 +380,46 @@ func (p *prober) doProbe(msg string, client kubernetes.Interface, pr *probeResul
...
+		klog.V(5).Infof("%s: probe failed with error: %s. Will retry...", msg, err)
 	}
 
 	if (err == nil && pr.lastError != nil) || (err != nil && pr.lastError == nil) {
 		pr.resultRun = 0
 	}
 
+	// Check if err is unauthorized or is forbidden and if the failure threshold is not reach.
+	// In case its true try to fetch the secret again and refresh the prober
+	if (apierrors.IsForbidden(err) || apierrors.IsUnauthorized(err) || apierrors.IsInvalid(err)) && pr.resultRun < p.failureThreshold {
```
unmarshall, line 394: `doProbe` takes in a client and is not really responsible for fetching secrets and refreshing clients. That should be the responsibility of the caller. `doProbe` should be kept simple - given a client it does the probe and returns the error. The caller of `doProbe` inspects the error and if it is one of `IsForbidden` or `IsUnauthorized` or `IsInvalid` then it does not increment the `resultRun` (very bad name), refreshes the clients and skips this run of the probe. In the next run of the probe the new clients will be used. If you feel that that delay is too much then you can either retry calling `doProbe` or reduce the time interval between 2 probes.

pkg/scaler/prober.go
```diff
@@ -538,14 +592,18 @@ func (p *prober) scaleUp(ctx context.Context) error {
...
 }
 
-// Checks for a given resource considered for scale, if for the respecitve scale operations its dependent deployments are in desired state.
+// Checks for a given resource considered for scale, if for the respective scale operations its dependent deployments are in desired state.
 // If availableReplicas is not equal to desired then it fails the check and the scaling fo the parent is stopped
 func (p *prober) checkScaleRefDependsOn(ctx context.Context, prefix string, dependsOnScaleRefs []autoscalingapi.CrossVersionObjectReference, replicas int32, checkFn func(oReplicas, nReplicas int32) bool) bool {
+	// running this check immediately after scaling tends to fail as the parent resource might still be processing
+	// introduce a short delay to let the parent resource availability to reflect current state correctly
+	// TODO: We should replace this with a proper flag for DWD probe, this requires also to identify the currect default value.
+	time.Sleep(toDuration(pointer.Int32Ptr(10), 0))
```
unmarshall, line 601: Not clear on what is the need to add 10s. Call to `checkScaleRefDependsOn` happens before the actual call to scale is made in function `retry(msg, p.getScalingFunc, gr, s, replicas)`

pkg/scaler/scaler.go
```diff
@@ -89,7 +89,8 @@ func NewController(clientset kubernetes.Interface,
 	})
 	c.secretsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(new interface{}) {
-			klog.V(4).Infof("Secret Added\n")
+			newSecret := new.(*v1.Secret)
```
unmarshall, line 92: if you are anyways casting the obj to `secret` then you can as well pass the same to `enqueueProbe` as well. This is handled in the refactoring that i am currently doing. So if you do not wish to include it as part of this PR then that would be fine.

unmarshall (commented):
Added comments

pkg/scaler/prober.go
ashwani2k: In the current common way of handling scaleDown and scaleUp we observed that during scaleDown without any delay. The scale down of dependent resources is not handled in the same reconciliation step. 
for e.g. if `KCM` is scale down and immediately after `MCM` is tried as it is dependent on `KCM` it checks if the available replicas for `KCM` is 0. As this is done immediately after `KCM` is scaled down, we observed that for most of the time it's not the case as the cache is not updated. So it fails and then skips scaling down `MCM` and then is retried in the next cycle. 

Eventually everything will be scaled down but it delays down everything by (no. of scaleDependents) * 30s to finally scale everything down. Hence on a safer side a 10s delay ensures that we scale down everything by (no. of scaledDependents) * 10s.

pkg/scaler/scaler.go
ashwani2k: I was thinking we can keep it this way for now, so that we can get some detailed logs for testing the secret rotation change if required. Otherwise we would need to wait for refactored one for it.  wdyt?

pkg/scaler/prober.go
unmarshall: ok, but 10s delay is a magic number and there is no guarantee that either the parent dependency is scaled down and the cache is also refreshed. The correct way of doing this is to have a pipeline of stages for scale down and scale up. Each step in the pipeline waits for its dependency to scale down. Once that is done in a timeout amount of time then it also scales down child dependencies, else it skips to the next cycle.
As also pointed out earlier, there is no strict parent <-> child relationship betwee `KCM` and `MCM`. If the external probe fails and resultRun > failureThreshold then both `KCM` and `MCM` can be stopped together. This again easily possible using a pipeline model as there can be sequential pipeline steps of parallel. For now you can go with waiting for 10s but please note that this is not always going to give you the desired outcome.

ashwani2k: Totally in line with your concern / proposal. 
 This is only for the current approach and keeping in mind that we wish to refactor the flow, I believe we can address it there.

pkg/scaler/scaler.go
unmarshall: ok sure.

pkg/scaler/prober.go
unmarshall: ok.

ashwani2k:
> 

Thanks Madhav, I agree we can make a total separation of concerns. 
The changes are made with the commit https://github.com/gardener/dependency-watchdog/pull/41/commits/15f476af3a05b03b7922b3866cf2a46df1687c51 and it also tested on a local seed/shoot setup.

ashwani2k:
Fixed test errors with missing arguments https://github.com/gardener/dependency-watchdog/compare/15f476af3a05b03b7922b3866cf2a46df1687c51..1d5b24b617cefdf54bef6493eddbec22aecf9108

unmarshall (approved):

PR #48, opened 2022-03-24
Annotation to ignore scaling added

himanshu-kun:
This PR adds support for ignoring scaling up dependents. If the dependent deployment has the annotation `dependency-watchdog.gardener.cloud/ignore-scaling` then it is ignored.

**Motivation:**

Currently DWD scales up the MCM, CA anytime they are scaled down. 
In Integration Tests(IT)  in MCM and CA , the tests scale-down MCM,CA and then run them locally and finally scale MCM, CA back up after getting done. Thats where the behavior of DWD described above causes issues.
This annotation will allow the tests to signal DWD not to scale up MCM, CA if they are scaled down due to a test. Finally when the tests are done they'll remove the annotation from the CA/MCM deployment.

Fixes #

**Release note**:

```noteworthy operator
A dependent's scaling up/down can be ignored by DWD now by adding the annotation `dependency-watchdog.gardener.cloud/ignore-scaling` to the deployment
```

### DIFF
--- a/pkg/scaler/prober.go
+++ b/pkg/scaler/prober.go
@@ -48,7 +48,7 @@ const (
 	defaultJitterSliding       = true
 
 	kindDeployment             = "Deployment"
-	ignoreScalingAnnotationKey = "gardener.cloud/ignore-scaling"
+	ignoreScalingAnnotationKey = "dependency-watchdog.gardener.cloud/ignore-scaling"
 )
 
 type prober struct {
--- a/pkg/scaler/prober.go
+++ b/pkg/scaler/prober.go
@@ -464,7 +464,7 @@ func (p *prober) updateClientsSecrets(pr *probeResult, msg string) {
 
 }
 
-func hasIgnoreScalingAnnotation(d *appsv1.Deployment) bool {
+func ignoreScalingDeployment(d *appsv1.Deployment) bool {
 	if val, ok := d.Annotations[ignoreScalingAnnotationKey]; ok {
 		return val == "true"
 	}
@@ -510,15 +510,16 @@ func (p *prober) scaleTo(parentContext context.Context, msg string, replicas int
 				klog.V(5).Infof("%s: replicas=%d: failed", prefix, replicas)
 				continue
 			} else {
+				if ignoreScalingDeployment(d) {
+					klog.V(4).Infof("%s: skipped because annotation %s present on deployment", prefix, ignoreScalingAnnotationKey)
+					continue
+				}
 				var specReplicas = int32(0)
 				if d.Spec.Replicas != nil {
 					specReplicas = *(d.Spec.Replicas)
 				}
 				if !checkFn(specReplicas, replicas) {
 					klog.V(4).Infof("%s: skipped because desired=%d and current=%d", prefix, replicas, specReplicas)
-					continue
-				} else if hasIgnoreScalingAnnotation(d) {
-					klog.V(4).Infof("%s: skipped because ignore scaling annotation present on deployment", prefix)
 					continue
 				}
 			}
@@ -665,7 +666,7 @@ func (p *prober) checkScaleRefDependsOn(ctx context.Context, prefix string, depe
 					return false
 				}
 				var availableReplicas = int32(0)
-				availableReplicas = d.Status.AvailableReplicas //check if available replicas is as desired
+				availableReplicas = d.Status.AvailableReplicas // check if available replicas is as desired
 				if !checkFn(availableReplicas, replicas) {
 					klog.V(4).Infof("%s: check for dependent %s succeeded as desired=%d and available=%d", prefix, d.Name, replicas, availableReplicas)
 					return true // can continue with scale operation of the parent
--- a/pkg/scaler/prober.go
+++ b/pkg/scaler/prober.go
@@ -47,7 +47,8 @@ const (
 	defaultJitterMaxFactor     = 0.2
 	defaultJitterSliding       = true
 
-	kindDeployment = "Deployment"
+	kindDeployment             = "Deployment"
+	ignoreScalingAnnotationKey = "gardener.cloud/ignore-scaling"
 )
 
 type prober struct {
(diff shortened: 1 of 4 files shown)


pkg/scaler/prober.go
```diff
@@ -47,7 +47,8 @@ const (
 	defaultJitterMaxFactor     = 0.2
 	defaultJitterSliding       = true
 
-	kindDeployment = "Deployment"
+	kindDeployment             = "Deployment"
+	ignoreScalingAnnotationKey = "gardener.cloud/ignore-scaling"
```
ashwani2k, line 51: Although I checked the annotation is not used anywhere in gardener org code base, however I was wondering if it is too generic to differentiate it's usage. 
Shall we add something to distinguish it from scaling operation done via other operations.

pkg/scaler/prober.go
```diff
@@ -508,6 +517,9 @@ func (p *prober) scaleTo(parentContext context.Context, msg string, replicas int
 				if !checkFn(specReplicas, replicas) {
 					klog.V(4).Infof("%s: skipped because desired=%d and current=%d", prefix, replicas, specReplicas)
 					continue
+				} else if hasIgnoreScalingAnnotation(d) {
+					klog.V(4).Infof("%s: skipped because ignore scaling annotation present on deployment", prefix)
```
ashwani2k, line 521: This will be applicable for both scaleUp and scaleDown operations. 
So what happens if - 
1. DWD brings down the MCM or other deployment as a result of the external probe failing and then someone applies this annotation to that resource. 
2. What if you wanted to only do it while scaling down but when scaling up or vice-versa.

ashwani2k (commented):
Thanks for the PR @himanshu-kun!

pkg/scaler/prober.go
himanshu-kun: how about `test.gardener.cloud/ignore-scaling-during-IT`

pkg/scaler/prober.go
himanshu-kun: yes this annotation doesn't restrict to scale-up ignoring only. The only `someone` to be expected to apply it IT. The IT after being done will scale up the deployment and remove the annotation, so its won't happen that MCM/CA won't come up.

ashwani2k: I was thinking for your integration test you would like to ensure that DWD doesn't scale up CA immediately if the annotation is applied. Why don't we deploy the config for DWD probe with the following 2 options -- 

1.  Remove `cluster-autoscalar` from dependentScales  in which case its ignored.  (this can be then restored to the default configs post tests. 

```
  dep-config.yaml: |
    namespace: ""
    probes:
    - dependantScales:
      - replicas: null
        scaleRef:
          apiVersion: apps/v1
          kind: Deployment
          name: kube-controller-manager
        scaleUpDelaySeconds: 120
      - replicas: null
        scaleRef:
          apiVersion: apps/v1
          kind: Deployment
          name: machine-controller-manager
        scaleRefDependsOn:
        - apiVersion: apps/v1
          kind: Deployment
          name: kube-controller-manager
        scaleUpDelaySeconds: 60
```

or 
2. you can a scaleUpDelay for CA to ensure that it is after your tests are passed. (this can be a permanent config for a given landscape as eventually CA will be scaled back and its not production critical. 
```
  dep-config.yaml: |
    namespace: ""
    probes:
    - dependantScales:
      - replicas: null
        scaleRef:
          apiVersion: apps/v1
          kind: Deployment
          name: kube-controller-manager
        scaleUpDelaySeconds: 120
      - replicas: null
        scaleRef:
          apiVersion: apps/v1
          kind: Deployment
          name: machine-controller-manager
        scaleRefDependsOn:
        - apiVersion: apps/v1
          kind: Deployment
          name: kube-controller-manager
        scaleUpDelaySeconds: 60
      - replicas: null
        scaleRef:
          apiVersion: apps/v1
          kind: Deployment
          name: cluster-autoscaler
        scaleUpDelaySeconds: 300
        scaleRefDependsOn:
        - apiVersion: apps/v1
          kind: Deployment
          name: machine-controller-manager
```

cc @unmarshall

pkg/scaler/prober.go
unmarshall: Having a test annotation is not such a great idea because now DWD needs to react to test annotations in its code. There are therefore 3 approaches: 

1. We can follow a similar approach to what a `gardener-resource-manager` does. It looks for  `resources.gardener.cloud/ignore=true` annotation on `ManagedResources` and if any such annotation is present then that resource will not be reconciled.
2. As you suggested, add an annotation to `DWD` to ignore scaling via annotation `gardener.cloud/ignore-scaling`
3. Update the configmap and add an additional property against `CA` to ignore scaling.

Let us analyze the approaches a bit:

**Approach-1:** 
This would mean that `cluster-autoscaler` resource will now have an additional annotation. Assumption is that the only component that will react to this annotation will be `DWD`. To ensure that change in the annotation is reflected in its informer cache, it could use the `initial-delay` field in the configuration for any scaling (scaling up or scaling down) operation. 

**Approach-2:** 
Adding an annotation `gardener.cloud/ignore-scaling` is ambiguous. As this annotation does not reflect on scaling of which resource should be avoided. Ideally if you only wish to prevent scaling of `CA` then you should be using a more specific annotation - example:  `resources.gardener.cloud/cluster-autoscaler/ignore-scaling`. 

**Approach-3:** 
Dynamically change the config map that is used by `DWD` to add additional field `ignore-scaling` with value `true` or any truthy value. Since this is a namespaced resource it is also an acceptable solution but it will involve more work, since you change the mounted config-map, DWD now needs to be capable of auto-reload the configuration and ensure that all probes that are started also are using the newly parsed configuration. This is a LOT of work.

pkg/scaler/prober.go
ashwani2k: I'm fine with keep  this as is for now. 
We can rethink on the solution when we refactor to see if there can be a better generic approach to handle such scenarios. 

/lgtm

ashwani2k (approved):
Looks good for the proposed behaviour we wish to achieve. However we should relook at it as part of refactor initiative to handle such cases in a more generic way.

ashwani2k:
/hold 

As there are some discussions still ongoing.

PR #52, opened 2022-04-29
Fixed typo in log message

ScheererJ:
The pull request fixes a minor typo in the log message of the dependency watchdog probe code.

None.

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/scaler/prober.go
+++ b/pkg/scaler/prober.go
@@ -424,7 +424,7 @@ func (p *prober) handleError(pr *probeResult, err error, msg string) {
 	if pr.lastError != nil {
 		klog.Errorf("%s: Probe finished with error %s for resultRun:%d", msg, pr.lastError.Error(), pr.resultRun)
 	} else {
-		klog.V(4).Infof("%s: Probe finished successfully for rusultRun:%d", msg, pr.resultRun)
+		klog.V(4).Infof("%s: Probe finished successfully for resultRun:%d", msg, pr.resultRun)
 	}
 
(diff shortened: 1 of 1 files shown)


ashwani2k:
Thanks @ScheererJ  for the PR. 
We are in works to refactor the DWD as we were not happy with not just the way some of the terms are defined like `resultRun` which doesn't convey anything meaningful, but we also intend to make the behavior more deterministic as we observed with race conditions spurring with recently introduced secret rotation in gardener as well as handling of meltdown scenarios where it wasn't sure how long should one wait for a dependent component to scale down. 

@unmarshall  is currently in the process of publishing the new approach to the community which should be out soon. The implementation for the same is already underworks.  So i'll rather close this in lieu of overhauling,  wdyt?

@rishabh-11  may be you can run a spell-check on the new code base to verify if such typos still exist in the new codebase.

unmarshall:
Yes i agree. We will double check all logs that DWD emits to ensure that it clearly depicts the message without spelling mistakes.
@ScheererJ i will publish the new approach soon. Implementation is also going at a good pace. Once we finish proper testing we will release the new version of DWD prober initially which will a lot of improvements. This is a very small change and if you still wish to merge this change i don't have any reasons to block it :)

ScheererJ:
@ashwani2k & @unmarshall: Thanks for the heads-up. Feel free to close or merge this pull request at your convenience. It is certainly not very important, especially if the dependency watchdog is anyway rewritten.

PR #56, opened 2022-05-27
fix for issue #55

### LINKED ISSUE #55: [BUG] DWD wrongfully scales down resources when api-server takes more than 10s to respond
**Describe the bug:**
Currently the DWD probe has a timeout of 10s which at times fails when there is request latency of higher than 10s. 
In such cases the api-server is reachable and kubelet is also able to update it's heartbeat, however DWD assumes that api-server is not reachable and bring the KCM/MCM/CA down. 

**Expected behaviour:**
DWD should honor the same timeout as set for kubelet. 

Create load on the kube-apiserver with high request load resulting latency values beyond 10s to respond. 
This should eventually fail the external probe and will eventually bring down the components although the api-server is reachable from the node albeit slow. 

**Logs:**

rishabh-11:
Adjust the probe timeout to be in sync with the kubelet timeout to update its lease.
Fixes #55 

**Release note**:

```bug operator
Increased DWD probe timeout to be aligned with the kubelet timeout.
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM alpine:3.15
+FROM alpine:3.15.4
 
 RUN apk add --update bash curl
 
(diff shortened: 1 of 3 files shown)


ashwani2k (approved):
Thanks @rishabh-11 for fixing this. 

/lgtm

PR #62, opened 2022-08-01
Fix for Issue#61

### LINKED ISSUE #61: DWD incorrectly scales down dependent services
**Describe the bug:**
DWD periodically attempts to probe the Kube ApiServer via a cluster-DNS endpoint (which directly calls the service) and an internal (also used by kubelet) endpoint (which is currently proxied via Envoy proxy). If the external probe fails for a configurable number of times it assumes that the Kube ApiServer is no longer reachable and to prevent a meltdown it scales down Kube Controller Manager, Machine Controller Manager and Cluster Autoscaler.

It has been recently observed that scaling down happens also in cases where the Kube ApiServer is reachable from the kubelet. However, DWD's probes continue to fail. On inspecting the TCP dump it was clear that the `client-go` and the underneath rest client re-uses the TCP connection as we could not see any `TCP SYN` packet being sent by the client for a long time. By default every 30 seconds a Keep-Alive packet will be sent. If the TCP connection is broken then due to KeepAlive, rest client will continue to use this connection and thereby causing failures of external probe.

The issue would not occur if a `TCP RST` packet is sent back to the client. However it is observed that especially on AWS no RST packet was ever sent thereby leading to no re-establish of TCP connection.

**Expected behaviour:**
Since probes are critical to up-time of control plane components like Kube Controller Manager, Machine Controller Manager and Cluster Autoscaler, it should be ensured that probe does not fail due to broken TCP connection.

**Logs:**

unmarshall:
Since DWD has no control over proper sending of TCP RST packets in case of a broken TCP connection, `KeepAlive=true` default configuration for the rest client causes re-use of the broken TCP connections which results in continuous failure of internal probes. This results in an incorrect scaling down of critical control plane components.

To prevent this the fix forcefully creates a new TCP connection for every internal and external probe. This is done by setting `DisableKeepAlive = `true` on the transport.

Fixes #61 

**Release note**:

DWD client shall no longer use long running TCP connections when attempting to probe Kube ApiServer via internal endpoint.
```improvement operator

```

### DIFF
--- a/pkg/scaler/httputil.go
+++ b/pkg/scaler/httputil.go
@@ -1,0 +1,38 @@
+package scaler
+
+import (
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+	"net/http"
+)
+
+type transportWrapper struct {
+	Transport http.RoundTripper
+}
+
+func (rt *transportWrapper) RoundTrip(req *http.Request) (*http.Response, error) {
+	klog.Infof("(transportWrapper)(RoundTrip) request: %v", req.URL.String())
+	return rt.Transport.RoundTrip(req)
+}
+
+func DisableKeepAlive(config *rest.Config) error {
+	transport, err := createTransportWithDisableKeepAlive(config)
+	if err != nil {
+		return err
+	}
+	config.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return &transportWrapper{Transport: transport}
+	})
+	return nil
+}
+
+func createTransportWithDisableKeepAlive(config *rest.Config) (*http.Transport, error) {
+	tlsConfig, err := rest.TLSConfigFor(config)
+	if err != nil {
+		return nil, err
+	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DisableKeepAlives = true
+	transport.TLSClientConfig = tlsConfig
+	return transport, nil
+}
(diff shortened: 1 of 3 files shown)


pkg/scaler/httputil.go
```diff
@@ -0,0 +1,39 @@
...
+	"k8s.io/klog"
+	"net/http"
+)
+
+type transportWrapper struct {
+	Transport http.RoundTripper
+}
+
+func (rt *transportWrapper) RoundTrip(req *http.Request) (*http.Response, error) {
+	klog.Infof("(transportWrapper)(RoundTrip) request: %v", req.URL.String())
```
ashwani2k, line 14: Should this be at a higher log level ?

ashwani2k (commented):
Thanks @unmarshall. 
Looks good to me. I've put one nitpick. 

Also as v2 of DWD is in works, so when we include this behaviour there,  if we can add a few words as comment on the usage of **function** `DisableKeepAlive` that will be good to understand why this is introduced. Also we can state the options checked and current limitations with `client-go`. 
This request is only for v2.

pkg/scaler/httputil.go
unmarshall: I changed it for debugging but you are right, this log should only be printed when log level is changed. Will change it to `klog.V(5).Infof` instead.

ashwani2k: Thanks for changing it. 
/lgtm

ashwani2k (approved):
Thanks for doing the changes. 
/lgtm

shreyas-s-rao (approved):
@unmarshall Thanks for fix as well as the revendoring.
/lgtm

PR #65, opened 2022-10-13
Make rules updated/introduced

himanshu-kun:
Introduced make rule `check-vulnerabilities` and `format`

Fixes #

**Release note**:

```improvement developer
`format` and `check-vulnerabilites` make targets introduced
```


unmarshall:
Changes are fine but a lot of libs are being pulled into vendor directory. Why is that the case?

himanshu-kun:
I have added `github.com/gardener/gardener/hack` as a required package in `hack/tools.go`. This is done so that the scripts can be vendored in.
So the dependencies listed in `github.com/gardener/gardener/hack` have also come in dwd.

cmd/probe.go
```diff
@@ -103,7 +103,12 @@ func runProbe(cmd *cobra.Command, args []string) {
 	leaderElectionClient := kubernetes.NewForConfigOrDie(rest.AddUserAgent(config, "dependency-watchdog-election"))
 	recorder := createRecorder(leaderElectionClient)
 	run := func(ctx context.Context) {
-		go serveMetrics()
+		go func() {
+			err := serveMetrics()
+			if err != nil {
+				klog.Fatalf("Can't serve metrics: %s", err)
```
unmarshall, line 109: Should this kill the container? Metrics are good to have and not must to have, right?

cmd/root.go
```diff
@@ -157,7 +157,12 @@ func runRoot(cmd *cobra.Command, args []string) {
 	leaderElectionClient := kubernetes.NewForConfigOrDie(rest.AddUserAgent(config, "dependency-watchdog-election"))
 	recorder := createRecorder(leaderElectionClient)
 	run := func(ctx context.Context) {
-		go serveMetrics()
+		go func() {
+			err := serveMetrics()
+			if err != nil {
+				klog.Fatalf("Can't serve metrics: %s", err)
```
unmarshall, line 163: Similar comment as the above.

unmarshall:
Looks good to me. Why is the unit test failing? Also build happens in concourse and test and verify build image happens in prow?

unmarshall (approved):
/lgtm

unmarshall:
/do-not-merge Have a look at why unit tests are failing

cmd/probe.go
himanshu-kun: okay will error on this

himanshu-kun:
PR updating `.ci/pipeline_definitions` ->  https://github.com/gardener/dependency-watchdog/pull/66

unmarshall (approved):

PR #71, opened 2022-11-21
`Optional` as a boolean field and not a pointer in prober api

rishabh-11:
This PR changes the prober API to make `optional` a boolean field and not a pointer.

Fixes #

**Release note**:

```improvement operator
NONE
```


internal/prober/testdata/config_missing_optional_values.yaml
```diff
@@ -5,7 +5,6 @@ dependentResourceInfos:
       kind: "Deployment"
       name: "kube-controller-manager"
       apiVersion: "apps/v1"
-    optional: false
```
unmarshall, line 8: The name of the file says it is for a config with missing optional values. A bit confusing as `optional` is both a field name in a struct and also it could also mean configuration where all optional values are absent.

api/prober/types.go
```diff
@@ -47,8 +47,8 @@ type Config struct {
 type DependentResourceInfo struct {
 	// Ref identifies a resource
 	Ref *autoscalingv1.CrossVersionObjectReference `json:"ref"`
-	// Optional should be true if this resource should be present. If the resource is optional then it should be false.
-	Optional *bool `json:"optional"`
+	// Optional should be false if this resource should be present. If the resource is optional then it should be true.
+	Optional bool `json:"optional,omitempty"`
```
unmarshall, line 51: omitempty should not be there as this field is not really optional. If one does not specify a value a zero value for this will be assumed which will be false. docstring also needs to reflect that very clearly.

rishabh-11: updated the docstring

unmarshall (approved):
/lgtm

Issue #76, opened 2023-01-24
[Improvement] Parameterize add-license-header makefile target 


unmarshall:
**Describe the bug:**
Year 2022 is hardcoded in [addlicenseheaders.sh](https://github.com/gardener/dependency-watchdog/blob/master/hack/addlicenseheaders.sh)

**Expected behaviour:**
Take year as a parameter when calling the make target.

unmarshall:
Fixed via #82

PR #82, opened 2023-04-24
For control-plane-as-a-service, no probes will be created

### LINKED ISSUE #80: Do not create probe for control-plane-as-a-service shoots

/area control-plane
/kind enhancement
/priority 3

[Gardener Issue#7635](https://github.com/gardener/gardener/issues/7635) introduces control plane as a service concept where number of workers will be 0 and number of control plane components will also be reduced. MCM and CA will not be deployed and there will also be no need to scale up/down KCM as there is no workload that is scheduled (as there are no nodes).

This enhancement optimises DWD and prevents creation of probes if for a Cluster the number of workers = 0. 

**Why is this needed**:

Currently DWD has a single configuration of dependent services for the prober which is applicable to all shoot control namespaces in the seed. This configuration is contained in a ConfigMap deployed in the `garden` namespace of the seed. For CPAAS (control plane as a service), there will be no deployments created for MCM and CA and there is no need to scale down KCM. Therefore for the prober there is nothing to do for these CPAAS namespaces.

unmarshall:
**How to categorize this PR?**
/kind enhancement
/area control-plane

Avoids creating probes for shoots which do not have any workers. 

Fixes #80 and #76

**Release note**:

```other operator
Probes will not be created for shoots with no workers.
```


controllers/cluster_controller.go
```diff
@@ -77,23 +77,34 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-			log.Info("Cluster migration is enabled, existing prober will be removed")
+			log.Info("Cluster migration is enabled, existing prober will be removed", "namespace", req.Namespace, "name", req.Name)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// if a shoot is created without any workers (this can only happen for control-plane-as-a-service use case), then if there is a probe registered then
+	// unregister the probe and return early. If there is no existing probe registered then return early.
+	if len(shoot.Spec.Provider.Workers) == 0 {
+		if r.ProberMgr.Unregister(req.Name) {
```
shafeeqes, line 104: Why do we need to do this? Can't we just use a predicate to not requeue the request itself if the shoot doesn't have workers? Then the probe is never created.

controllers/cluster_controller.go
```diff
@@ -64,7 +64,7 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// If the cluster is not found then any existing probes if present will be unregistered
 	if notFound {
 		if r.ProberMgr.Unregister(req.Name) {
-			log.Info("Cluster not found, existing prober will be removed")
+			log.Info("Cluster not found, existing prober will be removed", "namespace", req.Namespace, "name", req.Name)
```
shafeeqes: Doesn't the log contain these information already?

shafeeqes:
typo in PR name, `control-plane-as-a-service`

controllers/cluster_controller.go
unmarshall: Calling `Unregister` is essentially a NoOp in this case since no probe with the shoot ns will be found. However, it allows me to support any future use cases (with no additional cost) where one could transition from workerless to shoot with workers and vice versa. As DWD i really do not have to then care if something transitions are supported or not.

shafeeqes: Sure, but let's add a predicate as well and enqueue it only if there is a transition.

shafeeqes: I am suggesting the predicate like this:
```
if newHasWorkers{
  return true
}else if oldHasWorkers{
  return true
}
return false
```

hack/addlicenseheaders.sh
```diff
@@ -19,4 +19,13 @@ set -e
...
 # Uses the tool https://github.com/google/addlicense
-addlicense -v -c "SAP SE or an SAP affiliate company" -y 2022 -l apache -ignore "vendor/**" -ignore "**/*.md" -ignore "**/*.yaml" -ignore "**/Dockerfile" .
+YEAR=$1
+if [[ -z "$1" ]]; then
+  cat << EOF
+Unspecified 'YEAR' argument.
+Usage: addlicenceheaders.sh <YEAR>
+EOF
+  exit 1
+fi
```
acumino, line 29: Can we use date command here?

controllers/cluster_controller.go
unmarshall: Done

cmd/probercmd.go
```diff
@@ -18,7 +18,7 @@ import (
 	"flag"
 	"fmt"
 
-	"github.com/gardener/dependency-watchdog/controllers"
+	"github.com/gardener/dependency-watchdog/controllers/cluster"
 	"github.com/gardener/dependency-watchdog/internal/prober"
 	"github.com/gardener/dependency-watchdog/internal/util"
 	gardenextensions "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
```
shafeeqes, line 24: call it `extensionsv1alpha1` or `gardenerextensionsv1alpha1`

controllers/cluster/cluster_controller.go
```diff
@@ -153,7 +164,10 @@ func (r *ClusterReconciler) startProber(ctx context.Context, logger logr.Logger,
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&gardenerv1alpha1.Cluster{}).
+		For(
+			&gardenerv1alpha1.Cluster{},
+			ctrlbuilder.WithPredicates(workerLessShoot(mgr.GetLogger())),
```
shafeeqes, line 169: `mgr.GetLogger()` will not contain the controller name and kind. We should build the controller and use `c.GetLogger()` which is built with this info.
Ref: https://github.com/gardener/gardener/blob/b89462e43c1f56843d8d95bb1f459b1e0c0663b4/vendor/sigs.k8s.io/controller-runtime/pkg/builder/controller.go#L327-L336

cmd/probercmd.go
unmarshall: done

controllers/endpoint/endpoints_controller.go
```diff
@@ -57,20 +63,32 @@ func (r *EndpointReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	return c.Watch(
+		&source.Kind{Type: &v1.Endpoints{}},
+		&handler.EnqueueRequestForObject{},
+		predicate.And(
 			predicate.ResourceVersionChangedPredicate{},
-			MatchingEndpoints(r.WeederConfig.ServicesAndDependantSelectors), ReadyEndpoints(mgr.GetLogger()))).
-		WithOptions(controller.Options{MaxConcurrentReconciles: r.MaxConcurrentReconciles}).
-		Complete(r)
+			MatchingEndpoints(r.WeederConfig.ServicesAndDependantSelectors),
+			ReadyEndpoints(mgr.GetLogger()),
```
shafeeqes, line 91:
```suggestion
			ReadyEndpoints(c.GetLogger()),
```

controllers/cluster/cluster_controller.go
unmarshall: done

controllers/endpoint/endpoints_controller.go
unmarshall: This has been changed. PTAL.

controllers/cluster/cluster_controller.go
```diff
@@ -151,9 +167,16 @@ func (r *ClusterReconciler) startProber(ctx context.Context, logger logr.Logger,
...
+		controllerName,
+		mgr,
+		controller.Options{
+			Reconciler:              r,
+			MaxConcurrentReconciles: r.MaxConcurrentReconciles},
+	)
+	if err != nil {
+		return err
+	}
+	return c.Watch(&source.Kind{Type: &gardenerv1alpha1.Cluster{}}, &handler.EnqueueRequestForObject{}, workerLessShoot(c.GetLogger()))
```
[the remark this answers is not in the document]
unmarshall: done. PTAL

unmarshall:
@acumino, @shafeeqes I have now addressed all your review comments. PTAL.
@rishabh-11, @himanshu-kun please have a look at from DWD perspective and see if you want any change.

controllers/endpoint/endpoints_controller.go
```diff
@@ -26,12 +26,18 @@ import (
...
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// EndpointReconciler reconciles an Endpoints object
-type EndpointReconciler struct {
+const (
+	controllerName = "endpoint"
+)
```
shafeeqes, line 37: Nit:
```suggestion
const controllerName = "endpoint"
```

shafeeqes (commented):
Thanks for addressing my comments.
One minor comment.
Otherwise lgtm

controllers/endpoint/endpoints_controller.go
unmarshall: its done in both the controllers now. PTAL

shafeeqes (commented):
/lgtm

himanshu-kun (approved):
Thanks for the effort , the structure is much cleaner now!
Just a nit 
approving anyways

/lgtm

PR #94, opened 2023-12-13
Change DWD prober to check Node Leases

### LINKED ISSUE #85: DWD not working as expected on some infrastructures where traffic to LB stays in-cluster (gets rerouted internally)
Reminder for a finding recently. Details missing here.
cc @unmarshall @ashwani2k 

If memory serves, @ScheererJ and colleagues observed/noted that DWD will not actually reach out to/through the LB on all other infrastructures but AWS, which would make the test (API server reachable externally) not work as expected (reduced to an internal case, which cuts out problems like those for which we implemented the feature in the first place, e.g. a broken Istio ingress-gateway).

rishabh-11:
This PR removes the old logic of DWD prober doing two probes, internal and external. The internal probe still happens and is renamed to APIServerProbe, and a NodeLease probe is introduced, which replaces the external probe. It takes a look at the number of expired Node Leases and performs a scale-down if the threshold is crossed. Refer #85 to see why this is required.

Fixes #85 

**Release note**:

```breaking operator
DWD no longer depends on checking the reachability of Kube-API server via external load-balancer as kube-proxy path optimizations makes that impossible on infrastructures which have a static IP for the LB. Instead it now depends on node leases that are periodically renewed by each node running in the shoot cluster to determine its scaling action (scale-down/up).
```


internal/prober/prober_test.go
```diff
@@ -33,192 +42,159 @@ import (
...
 )
 
-type probeStatusEntry struct {
-	name                              string
-	err                               error
-	expectedInternalProbeSuccessCount int
-	expectedInternalProbeErrorCount   int
-	expectedExternalProbeSuccessCount int
-	expectedExternalProbeErrorCount   int
+type probeSimulatorEntry struct {
```
unmarshall, line 63: it is actually a test case struct entry. Therefore a more appropriate name would be `probeTestCase. Then in the tests you can say:
```go
testcases := []probeTestCase{}
```
But i find this approach a little sub-optimal for contributors. Reason is that for test case `TestAPIServerProbeFailure` you do not need most of the fields of the struct. Would it not be better to define the struct inline for every test case so that only what is required has to be set?

internal/prober/prober_test.go
```diff
@@ -33,192 +42,159 @@ import (
...
+	name                        string
+	leaseList                   *coordinationv1.LeaseList
+	mockShootClientCreatorError error
+	mockDiscoveryError          error
+	mockLeaseListError          error
+	mockScaleUpError            error
+	mockScaleDownError          error
+	minMockScaleUpCount         int
+	maxMockScaleUpCount         int
```
unmarshall, line 72: why call these with `Mock` in the name? This applies for most of the fields?

internal/prober/prober_test.go
```diff
@@ -33,192 +42,159 @@ import (
...
-	mki.EXPECT().Discovery().Return(mdi).AnyTimes()
-	mdi.EXPECT().ServerVersion().Return(nil, nil).AnyTimes()
-	runProberAndCheckStatus(t, 12*time.Millisecond, entry)
+	initializeMockInterfaces(entry)
+	p := NewProber(context.Background(), "default", config, mockScaler, mockShootClientCreator, proberTestLogger)
+	runProber(p, testProbeTimeout.Duration)
 }
 
-func runProberAndCheckStatus(t *testing.T, duration time.Duration, probeStatusEntry probeStatusEntry) {
+func createAndRunProber(t *testing.T, duration time.Duration, config *papi.Config, simulatorEntry probeSimulatorEntry) {
```
unmarshall, line 191: `simulatorEntry` is not used anywhere in this function.

internal/prober/prober_test.go
```diff
@@ -33,192 +42,159 @@ import (
...
-	mdi                                 *mockdiscovery.MockDiscoveryInterface
-	errNotIgnorable                     = errors.New("not Ignorable error")
-	internalProbeFailureBackoffDuration = metav1.Duration{Duration: time.Millisecond}
-	proberTestLogger                    = logr.Discard()
-	defaultProbeTimeout                 = metav1.Duration{Duration: 40 * time.Second}
+	ctrl                        *gomock.Controller
+	mockScaler                  *mockscaler.MockScaler
+	mockShootClientCreator      *mockprober.MockShootClientCreator
+	mockKubernetesInterface     *mockinterface.MockInterface
+	mockDiscoveryInterface      *mockdiscovery.MockDiscoveryInterface
```
unmarshall, line 52: defining these at the top might create problems if someone wishes to run these tests in parallel. I would rather create a struct containing all such mocked interfaces and pass it along after creating it for every test case.

internal/prober/prober_test.go
```diff
@@ -235,38 +211,43 @@ func runProber(p *Prober, d time.Duration) {
...
 
-func setupProberTest(t *testing.T) {
-	ctrl = gomock.NewController(t)
-	mds = mockscaler.NewMockScaler(ctrl)
-	msc = mockprober.NewMockShootClientCreator(ctrl)
-	clientBuilder = fake.NewClientBuilder()
-	fakeClient = clientBuilder.Build()
-	mki = mockinterface.NewMockInterface(ctrl)
-	mdi = mockdiscovery.NewMockDiscoveryInterface(ctrl)
+func initializeMockInterfaces(entry probeSimulatorEntry) {
```
unmarshall, line 224: can you rename this to `configureMocks` instead? 
If you decide to create these mocks for every test case then maybe you can club creation and configuration into a single function for a `ProberTestCase` and then call that function `createAndConfigureMocks`

internal/prober/prober_test.go
```diff
@@ -33,192 +42,159 @@ import (
...
-	table := []probeStatusEntry{
-		{"Scale Up Succeeds", nil, 1, 0, 1, 0},
-		{"Scale Up Fails", errors.New("scale Up failed"), 1, 0, 1, 0},
-	}
-
-	for _, probeStatusEntry := range table {
-		t.Run(probeStatusEntry.name, func(t *testing.T) {
-			setupProberTest(t)
-			config = createConfig(1, 1, metav1.Duration{Duration: 4 * time.Millisecond}, metav1.Duration{Duration: time.Microsecond}, 0.2)
+func TestSuccessfulProbesShouldRunScaleUp(t *testing.T) {
```
unmarshall, line 95: since you are using files, therefore you are limited to only 2 leases. Just create a `NodeLeaseCreator` which takes in a bunch of parameters allowing to to construct `coordinationv1.Lease`objects programmatically.

internal/prober/prober.go
```diff
@@ -26,37 +29,41 @@ import (
...
-	shootClientCreator  ShootClientCreator
-	internalProbeStatus probeStatus
-	externalProbeStatus probeStatus
-	ctx                 context.Context
-	cancelFn            context.CancelFunc
-	l                   logr.Logger
+	namespace          string
+	config             *papi.Config
+	scaler             dwdScaler.Scaler
+	shootClientCreator ShootClientCreator
```
unmarshall, line 54: In shootclient.go can you following camel casing for name of `shootclientCreator` struct. It should be renamed to `shootClientCreator` instead.

internal/prober/prober.go
```diff
@@ -87,36 +94,54 @@ func (p *Prober) Run() {
...
+// shouldPerformScaleUp returns true if the ratio of expired node leases to valid node leases is less than or equal to
+// the NodeLeaseFailureFraction set in the prober config
+func (p *Prober) shouldPerformScaleUp(candidateNodeLeases []coordinationv1.Lease) bool {
+	var validNodeLeaseCount, expiredNodeLeaseCount float64
+	for _, lease := range candidateNodeLeases {
+		if p.isLeaseExpired(lease) {
+			expiredNodeLeaseCount++
 		}
 	}
+	validNodeLeaseCount = float64(len(candidateNodeLeases))
```
unmarshall, line 144: a more apt name would be `candidateNodeLeaseCount` or get rid of this altogether. Instead just use 
```go
return expiredNodeLeaseCount/ float64(len(candidateNodeLeases)) < *p.config.NodeLeaseFailureFraction
```

internal/prober/prober.go
```diff
@@ -87,36 +94,54 @@ func (p *Prober) Run() {
...
-			p.l.Info("External probe is healthy, checking if scale up is already done or is still pending")
-			err := p.scaler.ScaleUp(ctx)
-			if err != nil {
-				p.l.Error(err, "Failed to scale up resources")
-			}
+		return
+	}
+}
+
+// shouldPerformScaleUp returns true if the ratio of expired node leases to valid node leases is less than or equal to
```
unmarshall, line 135: The description says `is less than or equal to` but the actual check does not check for equality but only less-than check is performed. So either fix the doc string or the implementation.

internal/prober/prober.go
```diff
@@ -87,36 +95,54 @@ func (p *Prober) Run() {
...
-	p.probeInternal(internalShootClient)
-	if p.internalProbeStatus.isHealthy(*p.config.SuccessThreshold) {
-		externalShootClient, err := p.setupProbeClient(ctx, p.namespace, p.config.ExternalKubeConfigSecretName)
+	if len(candidateNodeLeases) == 0 {
+		p.l.Info("No owned node leases are present in the cluster, skipping scaling operation")
+		return
+	}
+	if p.shouldPerformScaleUp(candidateNodeLeases) {
+		p.l.Info("Lease probe succeeded, performing scale up operation if required")
+		err := p.scaler.ScaleUp(ctx)
```
unmarshall, line 121: replace it with:
```go
if err = p.scaler.ScaleUp(ctx); err != nil {
   	p.l.Error(err, "Failed to scale up resources")
}
```

internal/prober/prober.go
```diff
@@ -87,36 +95,54 @@ func (p *Prober) Run() {
...
-		if p.externalProbeStatus.isUnhealthy(*p.config.FailureThreshold) {
-			p.l.Info("External probe is un-healthy, checking if scale down is already done or is still pending")
-			err := p.scaler.ScaleDown(ctx)
-			if err != nil {
-				p.l.Error(err, "Failed to scale down resources")
-			}
-			return
+	} else {
+		p.l.Info("Lease probe failed, performing scale down operation if required")
+		err := p.scaler.ScaleDown(ctx)
```
unmarshall, line 127: replace it with:
```go
if err = p.scaler.ScaleDown(ctx); err  != nil {
   p.l.Error(err, "Failed to scale down resources")
}
```

internal/prober/prober.go
```diff
@@ -176,3 +190,17 @@ func (p *Prober) doProbe(client kubernetes.Interface) error {
 	}
 	return nil
 }
+
+func (p *Prober) setBackOffIfThrottlingError(err error) {
+	if apierrors.IsTooManyRequests(err) {
+		p.l.V(4).Info("API server is throttled, backing off", "backOffDuration", backOffDurationForThrottledRequests.Seconds())
+		p.resetBackoff(backOffDurationForThrottledRequests)
```
elankath, line 197: the backoff handling doesn't appear correct, but since you have refactored this and new logic seems OK, will wait for next commit.

rishabh-11:
The testing of this PR was done on AWS and GCP. Attaching files for AWS and GCP logs, DWD config map and rootboy.txt (used for shutting down kubelet) :-

[AWS.md](https://github.com/gardener/dependency-watchdog/files/13991754/AWS.md)
[GCP.md](https://github.com/gardener/dependency-watchdog/files/13991755/GCP.md)
[dwd-cm.txt](https://github.com/gardener/dependency-watchdog/files/13991787/dwd-cm.txt)
[rootboy.txt](https://github.com/gardener/dependency-watchdog/files/13991792/rootboy.txt)

controllers/cluster/cluster_controller.go
```diff
@@ -113,8 +113,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
+	probeConfig := r.getEffectiveProbeConfig(shoot, log)
 	if canStartProber(shoot) {
-		r.startProber(ctx, log, req.Name)
+		r.startProber(ctx, probeConfig, log, req.Name)
```
elankath, line 118: (thanks for PR enhacements). There is already a `probeConfig` in the reconciler ? Why not override it after computing the effective `probeConfig`. Otherwise there is a stale `probeConfig` left dangling.

rishabh-11: The `probeConfig` of the reconciler is the default probe config generated from the config map used by DWD. We generate an effectiveProbeConfig because the `NodeMonitorGracePeriod` of KCM which is used in DWD as well (`kcmNodeMonitorGraceDuration`) is configurable via the shoot yaml. We should not change the config in the reconciler because otherwise the default value for `kcmNodeMonitorGraceDuration` will be lost.

unmarshall: To add what @rishabh-11 has already stated, probeConfig is one per seed and shared by all probers. However there is shoot specific configuration that we now need to consider which can override the nodeMonitorGracePeriod for a specific shoot and therefore a specific probe. Other shoots and their corresponding probes will continue to use the seed level prober config and only shoots where `nodeMonitorGracePeriod` has been explicitly set will take that value and thus have a different probe config.

elankath: Got it. Maybe we could call the field in the reconciler struct `baseProbeConfig` or `defaultProbeConfig`

rishabh-11: done

unmarshall (approved):

PR #112, opened 2024-04-19
Perform scale up if no candidate leases are present

### LINKED ISSUE #111: [Enhancement] If there are no node leases do a scale up of dependent resources

/area control-plane
/kind enhancement
/priority 2

There is an edge case where DWD scales down dependent resources due to `node lease probe` failure. If the expired node leases are cleaned up and no new lease gets created then DWD will skip any further scaling operation. This leaves the dependent resources in a scaled down state. This impacts the launch and registration of new nodes.

We should scale up dependent resources if they have been scaled down already and there are no node leases present.

rishabh-11:
This PR changes the code to performs scale up if no candidate leases are present

Fixes #111 

**Release note**:

```other operator
Perform scale up if no candidate leases are present
```

### DIFF
--- a/internal/prober/prober.go
+++ b/internal/prober/prober.go
@@ -108,12 +108,7 @@ func (p *Prober) probe(ctx context.Context) {
 	if err != nil {
 		return
 	}
-	if len(candidateNodeLeases) == 0 {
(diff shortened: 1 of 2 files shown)


unmarshall (approved):
/lgtm

Issue #123, opened 2024-05-13
Exclude terminating and Failed machines/nodes from DWD probe calculation of Failed Leases 


ashwani2k:
**How to categorize this issue?**

/area disaster-recovery
/area robustness
/kind enhancement
/priority 1

**What would you like to be added**:
Currently in DWD failed lease calculation all the nodes are considered to arrive at the `nodeLeaseFailureFraction`. 
This can be misleading for cases like:
1. When we have nodes which are in `Terminating` or `Failed` phase esp. if they will take the entire `machineDrainTimeout` due to PDB or other issues with eviction of pods. 
2. ~~Machines which are in crashLoopBackOff as they may never get created and including them in the count might also not be correct.~~

**Why is this needed**:
To avoid DWD to mistakenly initiate meltdown protection for clusters which are quick to hit the `nodeLeaseFailureFraction` if they are having prolonged occurrence of the above 2 phases for nodes/machines. 
Also observed in issue-live(4796)

PR #124, opened 2024-07-01
Prober should not consider `Failed` or `Terminating` Machines, `Unhealthy` Nodes. 

### LINKED ISSUE #110: [Enhancement] Consider node conditions apart from node leases to take more informed decisions for scale down

/area control-plane
/kind enhancement
/priority 3

DWD today checks if  the % of expired node leases is above a configured threshold then it will scale down the configured dependent resources (today that is KCM, MCM and CA). What it lacks is an ability to distinguish if the kubelet was unable to renew its lease due to network problems or was its own health or node's health.  If the kubelet or the node is unhealthy then DWD should not scale down MCM and KCM and let them collaborate in replacing the unhealthy node.

`Unhealthy Node`  is determined by looking at node conditions. Some of the node conditions are as below:

- DiskPressure
- KernelDeadlock
- ReadOnlyFileSystem
- KubeletUnavailable (this is set by node-problem-detector - currently this health checker plugin which sets this condition has not been enabled in g/g)

The above list is not comprehensive and therefore this list should be made configurable.

We therefore wish to introduce the following:
For all leases that are about to expire, check if the respective `Node` is present. If it is present then check  the `Conditions` on the `Node` object. If it is deduced that the node conditions indicate an unhealthy node then that node should not be counted in the set of nodes which have expired leases.

Let us take an example to explain this better:
- Consider a total of 10 nodes in a shoot cluster.
- Threshold number of expired leases: 60% = 6 nodes (If there are 6 or more nodes having leases that are about to expire then DWD should consider scale down of dependent resources).
- Let us assume that 7 nodes have leases that are about to expire.
- 2 out of these 7 nodes are unhealthy.

What happens today: Since 7 nodes have leases that are about to expire, all dependent resources are scaled down (this includes MCM  and KCM). This results in 2 unhealthy nodes not being replaced by MCM.

What we wish to have: 2 unhealthy nodes should be replaced by MCM but rest 5 should not be replaced.

rishabh-11:
This PR does the following:-
1. Prober will no longer consider `Failed` or `Terminating` Machines in candidate lease calculation.

2. Prober will no longer consider `Unhealthy` Nodes in candidate lease calculation. The criteria for a node to be `Unhealthy` is as follows:- 
    a.) The Shoot resource has a section to define node Conditions per worker to be checked by MCM for marking a machine `Unknown` See:- https://github.com/gardener/gardener/blob/98fb871eb6045f69715d9f8b39dd84f9637f1fa0/example/90-shoot.yaml#L138. If not specified DWD will use the default node conditions that MCM uses (See :- https://github.com/gardener/machine-controller-manager/blob/11c2983db70c3c13a7a95d51f9de9896e69d4599/pkg/util/provider/app/options/options.go#L60). 
    b.) If any of these node conditions is set to `True`, then DWD will consider the node `Unhealthy` and not consider it in candidate lease calculation.

3. Replace mocks with fake clients for unit tests.
4. If there is only one candidate lease present in the cluster, DWD probe will not perform any scaling operations.
5. Enabled Pprof endpoint to allow profiling of DWD. 

Fixes #110 #123 #126 

**Release note**:

```other operator
Replaced mocks with fake clients for unit tests
```
```feature operator
DWD prober no longer considers `Failed` or `Terminating` Machines in scaling decisions for the dependent resources.
```
```feature operator
DWD prober no longer considers `Unhealthy` Nodes in scaling decisions for the dependent resources.
```
```feature operator
DWD prober no longer performs scaling operations for clusters with only one candidate lease.
```


unmarshall (changes requested):
Just minor comments. Rest looks good

controllers/endpoint/endpoints_controller_test.go
```diff
@@ -150,9 +150,10 @@ func testWeederDedicatedEnvTest(t *testing.T) {
 	for _, test := range tests {
 		ctx, cancelFn := context.WithCancel(context.Background())
 		testEnv, reconciler := setupWeederEnv(ctx, t, test.apiServerFlags)
-		ns := testutil.CreateTestNamespace(ctx, g, reconciler.Client, strings.ToLower(test.name))
+		testNs := testutil.GenerateRandomAlphanumericString(g, 4)
```
unmarshall, line 153: You do not need to use this utility function any longer. We should perhaps remove this now.
Use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String instead.
Change this at all places where this utility function is used.

internal/docs/e2e-tests.md
```diff
@@ -1,52 +1,85 @@
...
-* Another way of running DWD is to change the image of the DWD deployment at `imagevector/images.yaml` in the cloned gardener repo used for the setup. (This needs to be done after the script checks out the desired gardener version and before `make gardener-extensions-up` is run)
+
+* Check the role/clusterrole for dependency-watchdog-prober in the `garden` namespace of the seed worker plane and add a
+  `patch` verb for deployment/scale resource if not present.
+* Check the role/rolebinding in the shoot for the dependency-watchdog-probe related service account in the `kube-system`
+  namespace. Add rules for listing leases in the `kube-node-lease` namespace.
+* Scale down the DWD prober deployment in the garden namespace (in the shooted-seed) and start a local DWD process by
+  providing the prober config and the kubeconfig of the shooted seed as command line flags -
```
unmarshall, line 37: you need to add a new line for commands that are enclosed within ```bash ``` else the formatting is not going to be correct.

internal/docs/e2e-tests.md
```diff
@@ -1,52 +1,85 @@
...
+DWD API server probe leverages an `apiserver-probe-endpoint`(name might vary)  to connect to the shoot Kube API server.
+API server probe DNS record points to an `in-cluster` endpoint which is only reachable from within the cluster. For
+tests that are run locally by starting a DWD prober process, the api server probe endpoint will have to be changed. This
+can be done in the following way:
 
+> **NOTE:** <br/> Create a new apiserver probe endpoint which is then reachable from the locally running DWD prober
+> process. To do that you will have to the following:
+> * Create a new `DNSRecord` containing the new shoot Kube API server endpoint. This will create a provider specific
+    route (e.g. In case of AWS it will create a AWS-Route53 endpoint)
+> * To ensure that a call to this endpoint is routed to the Kube API server of the shoot do the following:
```
unmarshall, line 56: Do we need all this? Please revisit and change the docs.

unmarshall (changes requested):

internal/prober/prober_test.go
```diff
@@ -25,180 +35,631 @@ import (
...
-	maxScaleDownCount       int
+func initializeTestScheme() *runtime.Scheme {
+	seedClientScheme := *scheme.Scheme
+	_ = v1alpha1.AddToScheme(&seedClientScheme)
+	return &seedClientScheme
 }
 
 func TestAPIServerProbeFailure(t *testing.T) {
-	testCases := []probeTestCase{
-		{name: "Unknown error is returned by api server", discoveryError: errFoo},
```
unmarshall, line 74: you seems to have removed this test case.

internal/test/miscellaneous.go
```diff
@@ -0,0 +1,58 @@
...
+	g := NewWithT(t)
+	var err error
+	if _, err := os.Stat(file); errors.Is(err, os.ErrNotExist) {
+		t.Fatalf("%s does not exist. This should not have happened. Check testdata directory.\n", file)
+	}
+	g.Expect(err).ToNot(HaveOccurred(), "File at path %v should exist")
+}
+
+// GenerateRandomAlphanumericString generates a random alphanumeric string of the given length.
+func GenerateRandomAlphanumericString(g *WithT, length int) string {
```
unmarshall, line 53: this function is no longer required.

unmarshall (changes requested):

internal/prober/prober_test.go
rishabh-11: It is not needed. The other cases are sufficient.

unmarshall (approved):
/lgtm

PR #137, opened 2025-04-07
Upgraded golang and dependencies

### LINKED ISSUE #60: CI should not build binary
After https://github.com/gardener/dependency-watchdog/pull/57 binary for DWD is built into the docker file https://github.com/gardener/dependency-watchdog/blob/606ac157ade6e1e85acaafe0ee2cc6b544c3166c/build/Dockerfile#L10

There is no need for CI to build binary again here.
https://github.com/gardener/dependency-watchdog/blob/606ac157ade6e1e85acaafe0ee2cc6b544c3166c/.ci/pipeline_definitions#L33-L35.

/cc @timuthy

### LINKED ISSUE #136: Upgrade golang and g/g dependency versions

/area dev-productivity
/kind enhancement
/priority 3

* Upgrade golang and gardener/gardener version
* Upgrade golangci-lint to v2

unmarshall:
This PR will do the following:

- Upgrades golang to 1.24
- Upgrades golangci-lint to v2
- Upgrades g/g to now use v1.115.2 (cannot use v1.116.0 due to https://github.com/gardener/gardener/pull/11819 not yet released as a patch)
- Corrected linting errors reported by the additional linters that got added.
- Updated the pipeline to use golang 1.24.1

Fixes #136

**Release note**:

```other operator
Upgraded to golang version=1.24, gardener/gardener version=1.115.2 and golangci-lint version = v2.0.2
```


gardener-robot-ci-2:
This PR proposes changes that would break the pipeline definition:
```
dependency-watchdog-pipeline: Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/concourse/replicator.py", line 141, in render
    definition_descriptor = self._render(definition_descriptor)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/concourse/replicator.py", line 185, in _render
    'definition': factory.create_pipeline_definition(),
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/concourse/factory.py", line 88, in create_pipeline_definition
    self._apply_traits(variant)
  File "/usr/lib/python3.12/site-packages/concourse/factory.py", line 177, in _apply_traits
    transformer.process_pipeline_args(pipeline_def)
  File "/usr/lib/python3.12/site-packages/concourse/model/traits/publish.py", line 750, in process_pipeline_args
    input_step = pipeline_args.step(input_step_name)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/concourse/model/job.py", line 238, in step
    raise ValueError(f'no such step: {name=}')
ValueError: no such step: name='build'

```

controllers/endpoint/endpoints_controller_test.go
```diff
@@ -177,7 +177,7 @@ func testWeederDedicatedEnvTest(t *testing.T) {
 func testOnlyCLBFPodDeletion(ctx context.Context, _ context.CancelFunc, g *WithT, reconciler *Reconciler, namespace string) {
 	createEp(ctx, g, reconciler, namespace, true)
 	pC := newPod(crashingPod, namespace, "node-0", correctLabels)
-	pH := newPod(healthyPod, namespace, "node-0", correctLabels)
+	pH := newPod(healthyPod, namespace, "node-1", correctLabels)
```
ashwani2k, line 180: It's not clear when I read this test as it use acronym `CLBF` without explaining what it means. 
Also we move to a new node for healthy pod, but then we should update the description above of the test to highlight this change.

controllers/endpoint/endpoints_controller.go
```diff
@@ -26,7 +26,7 @@ const controllerName = "endpoint"
 
 // Reconciler EndpointReconciler reconciles an Endpoints object
 type Reconciler struct {
-	client.Client
+	Client                  client.Client
```
ashwani2k, line 29: What's the benefit here for explicit declaration as the change is not reflecting elsewhere, is it just for readability?

ashwani2k (commented):
Thanks for the change, looks good to me, just some question for my own understanding on the usage or the best practice here.

controllers/endpoint/endpoints_controller.go
unmarshall: It is a good practice to define an explicit field for `Client`. If you do not do it then 2 things happen:
* golanglint issues warnings
* Semantically methods defined on the Client are for the Client and not on the reconciler. But if you embed `Client` then you can then call things like `reconciler.Get` that semantically does not make much sense. therefore you make it explicit.

controllers/endpoint/endpoints_controller_test.go
unmarshall: This was an existing method name that was not changed But you are right that we can have a better name. CLBF means crash loopbackoff

PR #140, opened 2025-05-01
Added meltdown protection annotation to resources managed by DWD

### LINKED ISSUE #135: Annotate DWD scaled resource to indicate DWD is actively managing replicas for this k8s resource

/area control-plane
/kind enhancement
/priority 2

When DWD scales-in dependent components like KCM/MCM/CA then it should annotate them with `dependency-watchdog.gardener.cloud/scaled-in` or `dependency-watchdog.gardener.cloud/meltdown-protection-activated`

**Why is this needed**:

It has been observed in productive clusters (e.g. live issue#6569) that decisions taken by DWD to scale down KCM/MCM/CA components can be partially reverted (only KCM replica change will be honoured). 

In gardener/gardener today only change in KCM's replicas are preserved. See [here](https://github.com/gardener/gardener/blob/9c10daba46f2d556a1de14f574e644ecaecc1ca3/pkg/gardenlet/operation/botanist/controlplane.go#L26).  However other control-plane components like MCM/CA do not have such checks. This causes MCM/CA to potentially scale up during the next shoot reconciliation even after DWD scaled them down.

Adding an annotation indicating that DWD has explicitly scaled-in the replicas for some of the control-plane components will help gardenlet decide during the reconciliation of the shoot if it needs to revert the replicas to what it desired. If this annotation is seen it can skip scaling-out these components.

ashwani2k:
This PR ensures that during the scale down operation DWD annotates the resources with annotation `dependency-watchdog.gardener.cloud/meltdown-protection-active`. This ensures that during shoot reconciliation the resources marked with this annotation are ignore from scaling up. 

The annotation shall be removed once DWD brings back up the components it had scaled down during meltdown protection period.

Fixes #135 

**Release note**:

```feature operator
A new annotation `dependency-watchdog.gardener.cloud/meltdown-protection-active` is now put on every resource scaled down by DWD to ensure that it is not acted upon during shoot reconciliation
```


internal/prober/scaler/scale.go
```diff
@@ -25,6 +25,8 @@ const (
 	// replicasAnnotationKey is the key for an annotation whose value captures the current spec.replicas prior to scale down for that resource.
 	// This is used when DWD attempts to restore the state of the resource it scale down.
 	replicasAnnotationKey = "dependency-watchdog.gardener.cloud/replicas"
+	// activeAnnotation captures if the DWD is active on a resource so that shoot reconciliation can ignore this resource from scaling up.
+	scaledDownAnnotationKey = "dependency-watchdog.gardener.cloud/scaled-down"
```
unmarshall, line 29: Is it not sufficient to indicate that DWD is active on a specific resource? Do we need to specify that the operation performed by DWD was `scaled-down`?
Perhaps we can use a more semantic name like `dependency-watchdog.gardener.cloud/meltdown-protection-active`. WDYT?

internal/prober/scaler/scale.go
```diff
@@ -162,6 +164,16 @@ func (r *resScaler) updateResourceAndScale(ctx context.Context, scaleSubRes *aut
 	}
 	if _, err = r.scaler.Update(childCtx, *gr, scaleSubRes, metav1.UpdateOptions{}); err != nil {
 		return err
+	} else {
```
unmarshall, line 167: There is no need for `else`. If there is an error during update of `Scale` subresource then the error is returned. IF there is no error then annotation is updated if the operation is scaleUp.

So this can be simplified for better code readability to:
```go
if _, err = r.scaler.Update(childCtx, *gr, scaleSubRes, metav1.UpdateOptions{}); err != nil {
  return err
}
if r.resourceInfo.operation == scaleUp {
  // rest of the code
}

```

internal/prober/scaler/scaler_test.go
```diff
@@ -428,12 +428,22 @@ func checkIfDeploymentReady(namespace, name string, replicas int32) bool {
 	return true
 }
 
-func checkScaleSuccess(g *WithT, opType operation, namespace, name string, expectedSpecReplicas int32) {
+func checkScaleSuccess(g *WithT, opType operation, namespace, name string, initialReplicas, expectedSpecReplicas int32) {
 	deploy := matchSpecReplicas(g, namespace, name, expectedSpecReplicas)
 	if opType == scaleUp {
-		g.Expect(deploy.Status.ReadyReplicas).To(BeNumerically(">=", 1))
```
unmarshall, line 434: Why was this removed?

internal/prober/scaler/scaler_test.go
```diff
@@ -428,12 +428,22 @@ func checkIfDeploymentReady(namespace, name string, replicas int32) bool {
...
+		g.Eventually(func() bool {
+			return checkIfDeploymentReady(namespace, name, expectedSpecReplicas)
+		}, 1*time.Minute, time.Second).Should(BeTrue())
+		g.Expect(deploy.ObjectMeta.Annotations).ToNot(HaveKey(scaledDownAnnotationKey))
 	} else {
-		g.Expect(deploy.Status.ReadyReplicas).To(BeNumerically("==", 0))
```
unmarshall, line 436: why was this removed?

unmarshall:
Thanks for raising this PR and ensuring that you have modified the unit tests as well. I have left a few review comments. PTAL.

internal/prober/scaler/scale.go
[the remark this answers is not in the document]
ashwani2k: Well makes sense, but this is an existing code and hence I didn't touch it as it was already reviewed. 
Anyways done with https://github.com/gardener/dependency-watchdog/pull/140/commits/0909963107f75efcd9bfa5d55df2907023df39a7

ashwani2k: Done with https://github.com/gardener/dependency-watchdog/pull/140/commits/0909963107f75efcd9bfa5d55df2907023df39a7, that was my bad.

ashwani2k (commented):
I've updated based on suggested.

internal/prober/scaler/scale.go
unmarshall: Ok my thinking was that once you add an annotation indicating that dwd is currently scaling this resource, then both scale-in and scale-out operations are automatically covered. Gardener looks at this annotation and knows that during its reconciliation it should not scale this resource yet since DWD is acting on this resource.

Also since this is now going to be used by gardener as well, it therefore becomes part of the API. Can you move this annotation to the API (api/prober/types.go) - there you can define a constant for this annotation and add proper doc comments.

internal/prober/scaler/scaler_test.go
unmarshall: ok, thanks

unmarshall (changes requested):

internal/prober/scaler/scale.go
ashwani2k: As suggested, adapted via https://github.com/gardener/dependency-watchdog/pull/140/commits/1c3a55ae5bc69bb9e9171e51bc0c017a4ae30fc9.

unmarshall (approved):

PR #213, opened 2026-06-19
Adjuster for Machine Creation Timeout - Basic Implementation

### LINKED ISSUE #201: Observe Machines Failing Creation and adjust shoot CA NodeGroup maxNodeProvisionTime

/area auto-scaling
/kind enhancement
/priority 2

Add DWD capability to observe `Machines` that regularly fail creation on shoot clusters of the seed and adjust shoot cluster-autoscaler NodeGroup `maxNodeProvisionTime` between the specified `workers[i].clusterAutoscaler.maxNodeProvisionTime` and `workers[i].clusterAutoscaler.maxNodeProvisionTimeUpperBound` in the shoot spec, so that the CA and the MCM increase the time given for machine provisioning before terminating the `Machine` and re-creating the same.

NOTE: Take learnings from other shoot clusters of the seed, so that each shoot does not need individual monitoring and adjustment of CA `maxNodeProvisionTime`. Exact mechanics of this will be detailed later.

Dependent backlog: https://github.com/gardener/machine-controller-manager/issues/1098 

**Why is this needed**:
Machines on slow, overloaded providers (CCloud) with the default machine creation timeout of 20m were stuck in creation flow for a long time due to regular terminations and re-creations. The objective is to let the cluster self-heal without manual intervention.

elankath:

Fixes #201

**Release note**:

```feature operator
Adjust effective machine-creation-timeout on MachineDeployment after breaching failure threshold
```


api/adjuster/.import-restrictions
```diff
@@ -0,0 +1,9 @@
+rules:
+  - selectorRegexp: (.+[.])?k8s[.]io
+    allowedPrefixes:
+      - k8s.io/api
+      - k8s.io/apimachinery
+  - selectorRegexp: github[.]com/gardener/dependency-watchdog
+    allowedPrefixes:
+    # should be self-contained and must not import any other dependency watchdog packages
+      - github.com/gardener/dependency-watchdog/api/prober
```
unmarshall: why is dependency on api/prober allowed?

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+const (
+	// AnnotationKeyEffectiveCreationTimeout is the annotation key set on the MachineDeployment that indicates
+	// the effective creation timeout for all Machine's belonging to this MachineDeployment. If specified, the value for this
+	// annotation takes precedence over the MachineDeployment.Spec.Template.Spec.MachineCreationTimeout.
+	AnnotationKeyEffectiveCreationTimeout = "node.machine.sapcloud.io/effective-creation-timeout"
+	// AnnotationKeyLastAdjustedEffectiveCreationTimeout is the annotation key set on the MachineDeployment that indicates
+	// the time that the effective creation timeout was last adjusted for this MachineDeployment.
+	AnnotationKeyLastAdjustedEffectiveCreationTimeout = "node.machine.sapcloud.io/last-adjusted-effective-creation-timeout"
```
unmarshall: The name `LastAdjustedEffectiveCreationTimeout` reads as "the previous value of the effective creation timeout" (a duration) where as the content is a timestamp (when the adjustment happened). So the doc string is correct but the name of the constant is misleading.

One option: `AnnotationKeyEffectiveCreationTimeoutLastAdjustedAt`

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+	// AnnotationKeyMachineJoinDuration is the annotation key set on the Machine that indicates the amount of time Machine
+	// took to join the cluster. The value is a Go Duration string.
+	AnnotationKeyMachineJoinDuration = "node.machine.sapcloud.io/machine-join-duration"
```
unmarshall: This constant is not used anywhere. What is the purpose of it?

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+	//DefaultCreationTimeoutGrowthFactor is the default value for growth of the effective-creation-timeout.
+	DefaultCreationTimeoutGrowthFactor = 2.0
```
unmarshall, line 23: I am not sure if doubling is the right growth factor - but you can check with Ashwani once. It should be based on some data that he might be privy to.

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+	InstanceType string
+	Zone         string
+}
+
+func (k ProvisionKey) String() string {
+	return fmt.Sprintf("(instanceType=%s,Zone=%s)", k.InstanceType, k.Zone)
+}
+
+// Controller is the facade exposed by the adjuster controller.
+type Controller interface {
```
unmarshall: Why is this part of the API?
This file is types.go and if you see the other 2 components `prober` and `weeder` then they do not define any `Controller` interface in their types and restrict defining types that are supposed to be used by the consumers.

If you wish to create this interface then you can move it directly under `controllers` and then also use that in both `cluster` and `endpoint` controllers.

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+	FailureThreshold *uint32 `json:"failureThreshold"`
+}
+
+// ProvisionKey identifies a group of machines with the same instance
+// type and availability zone for which provisioning statistics are tracked.
+//
+// Machines with the same instance type and availability zone share the same
+// provisioning characteristics, such as capacity availability, machine join
+// latency, and provisioning failure rates.
+type ProvisionKey struct {
```
unmarshall: Why is this part of the API?
Its only used by the controller.

cmd/adjustercmd.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+var (
+	// AdjusterCmd stores info about using the adjuster command
+	AdjusterCmd = &Command{
+		Name:      "adjuster",
+		UsageLine: "",
+		ShortDesc: "Checks Machine metrics/status and adjusts machine-creation-timeout on MachineDeployment's",
+		LongDesc: `For the seed cluster, it will start a watch on Machine objects, check to see if they have failed to join
+the shoot cluster within twice the machine-creation-timeout. If not, it will adjust the effective machine-creation-timeout
```
unmarshall, line 31:
```
check to see if they have failed to join the shoot cluster within twice the machine-creation-timeout
```
What does this mean?

```
This is adjusted upwards until 2h is reached
```
You cannot have the command description based on default values alone. If a different value is set via the Configuration then the command description is then invalid.

cmd/adjustercmd.go
```diff
@@ -0,0 +1,119 @@
...
+on the corresponding MachineDeployment with the same instance type and zone combination. This is adjusted upwards until
+2h is reached. If all Machines belonging to an instance type & zone combo have successfully joined the cluster in a window that is
```
unmarshall, line 33:
```If all Machines belonging to an instance type & zone combo have successfully joined the cluster in a window that is  twice the machine creation timeout
```
How do you ensure that? If there is a churn - where machines are getting scaled every few mins (within the window) then the predicate that all Machines joined the cluster (registered as Node) is not going to be true, right?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// LoadConfig reads the adjuster configuration from a file if specified, unmarshalls into instance of
+// [adjustapi.Config], fills default values for missing config fields using [FillConfigDefaults] and returns
+// populated object or an error.
+func LoadConfig(configPath string) (config *adjustapi.Config, err error) {
```
unmarshall, line 22: can you move the config related functions to a new file with the name `config.go`?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+	return config, nil
+}
+
+// ValidateConfig validates the given adjuster Configuration.
+func ValidateConfig(c *adjustapi.Config) error {
+	v := new(util.Validator)
+	if c.CreationTimeoutMax != nil {
+		v.MustNotBeZeroDuration("CreationTimeoutMax", *c.CreationTimeoutMax)
+	}
+	if c.CreationTimeoutGrowthFactor != nil {
```
unmarshall, line 45: how can this be nil?
If you are running FillConfigDefaults then there is no way this can be nil, right?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
+package adjuster
```
unmarshall, line 1: This file has become a dumping ground - usual problem with generic `helper.go` files.

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+		v.MustBeGreater("CreationTimeoutGrowthFactor", *c.CreationTimeoutGrowthFactor, 1.0)
+	}
+	if v.Error != nil {
+		return v.Error
+	}
+	return nil
+}
+
+// FillConfigDefaults fills default values for missing config entries inside the given [adjustapi.Config]
+func FillConfigDefaults(c *adjustapi.Config) {
```
unmarshall, line 55: any reason this is exposed?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+	FillConfigDefaults(config)
+	err = ValidateConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+// ValidateConfig validates the given adjuster Configuration.
+func ValidateConfig(c *adjustapi.Config) error {
```
unmarshall, line 40: any reason this is exposed?

controllers/adjuster/helper.go
unmarshall, line 22: any reason this is exposed?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+
+// FillConfigDefaults fills default values for missing config entries inside the given [adjustapi.Config]
+func FillConfigDefaults(c *adjustapi.Config) {
+	c.CreationTimeoutGrowthFactor = util.GetValOrDefault(c.CreationTimeoutGrowthFactor, adjustapi.DefaultCreationTimeoutGrowthFactor)
+	c.CreationTimeoutMax = util.GetValOrDefault(c.CreationTimeoutMax, metav1.Duration{Duration: adjustapi.DefaultCreationTimeoutMax})
+	c.FailureThreshold = util.GetValOrDefault(c.FailureThreshold, adjustapi.DefaultFailureThreshold)
+}
+
+// EventPredicate creates controller runtime [predicate.Predicate] for [machinev1alpha1.Machine] Updated event that satisfy [CanAcceptForAdjusterReconcile]
+func EventPredicate(logger logr.Logger) predicate.Predicate {
```
unmarshall, line 63: Perhaps it makes better sense to create  a new file `register.go` which should have predicates and SetupWithManager

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	machineNew, ok := objNew.(*machinev1alpha1.Machine)
+	if !ok {
+		return
+	}
+	oldPhase := machineOld.Status.CurrentStatus.Phase
+	currPhase := machineNew.Status.CurrentStatus.Phase
+
+	// TODO: BEGIN REMOVE ME {
+	if !strings.Contains(machineNew.Name, "i034796") {
```
unmarshall: please remove this

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+		GenericFunc: func(_ event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
+// CanAcceptForAdjusterReconcile returns true if
+//   - Machine moved from PENDING->FAILED phase
+//   - Machine moved from PENDING->RUNNING phase
+func CanAcceptForAdjusterReconcile(log logr.Logger, objOld, objNew client.Object) (accept bool) {
```
unmarshall, line 83: any reason to expose this function?

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+		return r.reconcileMachineJoin(ctx, m)
+	} else if r.isFirstSeenFailed(m) {
+		return r.reconcileMachineFail(ctx, m)
+	} else if m.DeletionTimestamp != nil {
+		r.state.clearDataForMachine(req.NamespacedName)
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *defaultController) isFistSeenPending(m *machinev1alpha1.Machine) bool {
```
unmarshall:
```suggestion
func (r *defaultController) isFirstSeenPending(m *machinev1alpha1.Machine) bool {
```

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+func (r *defaultController) isFirstSeenFailed(m *machinev1alpha1.Machine) bool {
+	return machineutils.IsMachineFailed(m) && !r.state.isFailRecorded(client.ObjectKeyFromObject(m))
+}
+func (r *defaultController) reconcileMachineCreate(ctx context.Context, m *machinev1alpha1.Machine) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	machineClass, err := r.getMachineClass(ctx, m)
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, nil
+	}
+	if machineClass == nil || err != nil {
```
unmarshall: how can `machineClass` be nil? If its a NotFound error then you are returning early. If you are here then machineClass should never be nil, right?

controllers/adjuster/adjuster.go
```diff
@@ -0,0 +1,342 @@
...
+	return machineutils.IsMachineFailed(m) && !r.state.isFailRecorded(client.ObjectKeyFromObject(m))
+}
+func (r *defaultController) reconcileMachineCreate(ctx context.Context, m *machinev1alpha1.Machine) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	machineClass, err := r.getMachineClass(ctx, m)
+	if apierrors.IsNotFound(err) {
+		return ctrl.Result{}, nil
+	}
+	if machineClass == nil || err != nil {
+		return ctrl.Result{}, nil
```
unmarshall: If there is an error getting the machineClass, should you not be logging it? This will ensure that you have a log stating that you got the event but could not record/update the internal state.

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+		newTimeout = maxTimeout
+	}
+	if newTimeout == currTimeout {
+		return 0
+	}
+	return newTimeout
+}
+
+// GetMachineDeploymentName gets the name of the MachineDeployment associated with this Machine
+func GetMachineDeploymentName(machine *machinev1alpha1.Machine) string {
```
unmarshall, line 192: is this really necessary? and why is it exposed?

controllers/adjuster/helper.go
unmarshall, line 1: for prober and weeder components all of this we have put under `internal`. The controller package has a dependency on the internal package. Is there a reason why we diverged from that?

controllers/adjuster/adjuster.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	deploymentName := GetMachineDeploymentName(m)
+	if deploymentName == "" {
```
unmarshall, line 95: can this ever happen?
If not and you still wish to handle it then there are inconsistencies. For example when you create a ProvisionKey then you assume `corev1.LabelTopologyZone` will be present. So for some labels you assume its present and for some you assume it could be absent. Can you explain the divergence?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+	} else {
+		log.V(5).Info("Machine skipped for adjuster reconcile", "name", machineNew.Name, "oldPhase", oldPhase, "currPhase", currPhase, "status.lastOperation", machineNew.Status.LastOperation)
+	}
+	return
+}
+
+// GetEffectiveCreationTimeoutOnMachine gets the effective creation timeout for this Machine object, first checking
+// the [adjustapi.AnnotationKeyEffectiveCreationTimeout], then falling back to machine spec and then falling back
+// to [adjustapi.StandardCreationTimeout].
+func GetEffectiveCreationTimeoutOnMachine(m *machinev1alpha1.Machine) (duration time.Duration, err error) {
```
unmarshall, line 116: Ideally all of this should have been there in `internal` package. If you insist to keep it here then there is no need to expose this.

api/adjuster/types.go
```diff
@@ -0,0 +1,73 @@
...
+	//DefaultCreationTimeoutGrowthFactor is the default value for growth of the effective-creation-timeout.
+	DefaultCreationTimeoutGrowthFactor = 2.0
+	//DefaultCreationTimeoutMax is the max limit beyond upto which the machine-creation-timeout can be adjusted
+	DefaultCreationTimeoutMax = 90 * time.Minute
+	// DefaultFailureThreshold is the default value of the threshold for number of Failed Machines beyond which
+	// the adjuster will revise the machine-creation-timeout.
+	DefaultFailureThreshold = 1
+	// StandardCreationTimeout is the standard machine creation timeout in gardener clusters if not overridden.
+	// See https://github.com/gardener/gardener/blob/e140ccc402b8732499cb804190fc4fe1ce82c078/example/90-shoot.yaml#L142
+	StandardCreationTimeout = 20 * time.Minute
```
unmarshall, line 36: Why is this not named `DefaultStandardCreationTimeout`?
Also in `Config` - i do not see this being overridden. I also do not see this value being read from Shoot resource.

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+		log.V(2).Info("Cannot reconcileMachineCreate - no 'name' (deployment name) on Machine", "machineName", m.Name)
+		return ctrl.Result{}, nil
+	}
+	pKey := adjustapi.ProvisionKey{
+		InstanceType: machineClass.NodeTemplate.InstanceType, // TODO: later this will be available as label on Machine object
+		Zone:         m.Spec.NodeTemplateSpec.Labels[corev1.LabelTopologyZone],
+	}
+	timeout, err := GetEffectiveCreationTimeoutOnMachine(m)
+	if err != nil {
+		log.Error(err, "Cannot reconcileMachineCreate")
```
unmarshall: message is very generic without contextual message/info.

controllers/adjuster/adjuster.go
```diff
@@ -0,0 +1,342 @@
...
+		InstanceType: machineClass.NodeTemplate.InstanceType, // TODO: later this will be available as label on Machine object
+		Zone:         m.Spec.NodeTemplateSpec.Labels[corev1.LabelTopologyZone],
+	}
+	timeout, err := GetEffectiveCreationTimeoutOnMachine(m)
+	if err != nil {
+		log.Error(err, "Cannot reconcileMachineCreate")
+		return ctrl.Result{}, nil
+	}
+	// TODO: Q: Is this acceptable ttl for map entry?
+	ttl := time.Duration(float64(timeout) * *r.config.CreationTimeoutGrowthFactor * 4)
```
unmarshall: shouldn't the TTL be a function of `CreationTimeoutMax` + some margin instead of a product that involves a growth factor?

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+}
+func (r *defaultController) isFirstJoin(m *machinev1alpha1.Machine) bool {
+	return hasJoined(m) && !r.state.isJoinRecorded(client.ObjectKeyFromObject(m))
+}
+func (r *defaultController) isFirstSeenFailed(m *machinev1alpha1.Machine) bool {
+	return machineutils.IsMachineFailed(m) && !r.state.isFailRecorded(client.ObjectKeyFromObject(m))
+}
+func (r *defaultController) reconcileMachineCreate(ctx context.Context, m *machinev1alpha1.Machine) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	machineClass, err := r.getMachineClass(ctx, m)
```
unmarshall: `machineClass` is only required to get the `InstanceType` right?
K8S has a stable label for instance-type: `node.kubernetes.io/instance-type` - is that not put on the Machine object today?

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+		return ctrl.Result{}, nil
+	}
+	// TODO: Q: Is this acceptable ttl for map entry?
+	ttl := time.Duration(float64(timeout) * *r.config.CreationTimeoutGrowthFactor * 4)
+	trackInfo := machineTrackInfo{
+		provisionKey:   pKey,
+		namespacedName: client.ObjectKeyFromObject(m),
+		ttl:            ttl,
+	}
+	createDuration := time.Since(m.CreationTimestamp.Time)
```
unmarshall: `createDuration` - is easy to misunderstand as how long the creation took.
I am not sure why do you even wish to capture this information and not capture just m.CreationTimestamp.

controllers/adjuster/controller.go
```diff
@@ -0,0 +1,342 @@
...
+	} else if r.isFirstSeenFailed(m) {
+		return r.reconcileMachineFail(ctx, m)
+	} else if m.DeletionTimestamp != nil {
+		r.state.clearDataForMachine(req.NamespacedName)
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *defaultController) isFistSeenPending(m *machinev1alpha1.Machine) bool {
+	return isPending(m) && !r.state.isRecorded(client.ObjectKeyFromObject(m))
```
unmarshall: For a machine that is still Pending, `machineTrackInfos` entries expire after `timeout × growthFactor × 4`
A machine that stays Pending longer than that loses its entry in the cache. There is no recovery - because the predicate will disallow another event for a Machine that is still in `Pending` because `oldPhase` will not be empty.

The same state can be reached when:
* `getMachineClass` errors
* `GetEffectiveCreationTimeoutOnMachine` errors
In the above 2 cases the `reconcileMachineCreate` will return without tracking the machine create leaving no entry into the cache. Then no new event will be permitted due to the event-predicate.

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+type state struct {
+	mu *sync.Mutex
+	// machineTrackInfos is an expiring map of machine namespaced name to machineTrackInfo
+	machineTrackInfos *cache.Expiring
+	// deploymentStats is an expiring map of MachineDeployment namespaced name to machineDeploymentStat
+	deploymentStats *cache.Expiring
+	// provisionKeyEntries is an expiring map of [adjustapi.ProvisionKey] to provisionKeyEntry
+	provisionKeyEntries *cache.Expiring
```
unmarshall: ProvisionKey is a combination of AZ + InstanceType - then why does `provisionKeyEntry` have a set of `deploymentNames`?
A Deployment is a combination of Az + instance type anyways?

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+		ttl:            trackInfo.ttl})
+	s.recordProvisionEntry(trackInfo.provisionKey, mcdNsName, deploymentStat.ttl)
+	return deploymentStat
+}
+
+func (s *state) trackMachineJoin(machineNsName types.NamespacedName, mcdNsName types.NamespacedName, joinDuration time.Duration) (trackInfo machineTrackInfo, deploymentStat machineDeploymentStat, ok bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	trackInfo, ok = s.getMachineTrackInfo(machineNsName)
+	if !ok {
```
unmarshall: if `s.getMachineTrackInfo` returns `!ok` then why do you return and not record this in `s.updateMachineDeploymentStat`? 
It is possible that you missed the create event and when you did get an event the oldPhase is not empty.
So you should reconstruct the ProvisionKey from the MachineClass (as you do in the create path), this will prevent a silent drop.

unmarshall (changes requested):
My review is not complete. But here are the initial set of comments

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (t machineTrackInfo) String() string {
+	return fmt.Sprintf("(provisionKey=%s,namespacedName=%s,joined=%t,failed=%t,ttl=%s)",
+		t.provisionKey, t.namespacedName, t.joined, t.failed, t.ttl)
+}
+
+// machineDeploymentStats stores aggregated observations for a MachineDeployment within a ttl.
+type machineDeploymentStat struct {
+	ttl            time.Duration
+	createDuration time.Duration // createDuration maximum
```
unmarshall: this field should be renamed to `maxCreateDuration` as it holds the max time across all machines of a single MCD.

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+	joinDuration   time.Duration // join duration maximum
```
unmarshall: this field should be renamed to `maxJoinDuration` as it holds the max time across all machines of a single MCD.

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,233 @@
...
+
+// GetMachineDeploymentName gets the name of the MachineDeployment associated with this Machine
+func GetMachineDeploymentName(machine *machinev1alpha1.Machine) string {
+	return machine.Labels["name"]
+}
+
+func combineDeploymentStat(existingData, data machineDeploymentStat) machineDeploymentStat {
+	data.createDuration = max(existingData.createDuration, data.createDuration)
+	data.joinDuration = max(existingData.joinDuration, data.joinDuration)
+	if data.joinCount > 0 {
```
unmarshall: I did not quite understand this logic of reseting either `joinCount` or `failCount`.
Lets say MCD has 10 replicas. MCM starts to create 10 machines, 5 fail and 5 succeed.
If the interleaving happens in a way where join event comes in the end then it flips all failCount to 0. So even if the `FailureThreshold` is set to 1 (default) and there are 5 fail machines - `reconcileMachineFail` will not trigger since join even resets the fail count. 

I think your current model is a `streak` model. So a single machine that fails `n` no of times then you wish to take an action. But here you are accumulation across machines for a single MCD. So i am confused on what is intended here.

The question that needs to be answered instead is:
```
Within a time window, how many machines of a given (zone + instanceType) combination failed to provision?
```
If this count exceeds a threshold then grow the creation timeout of all MCDs that use that combination.
So what you need instead is a `windowed count of failures per provision key with no cross-resetting`.

- When a machine fails increment failCount for the provision key (ensure dedup)
- TTL window handles recency - old failures are aged out.
- Joins mutates `maxJoinDuration` but do not reset `failCount`.
- Then the threshold check `failCount >= FailureThreshold` at the granularity of provision key.

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (s machineDeploymentStat) String() string {
+	return fmt.Sprintf("(joinCount=%d,failCount=%d,createDuration=%s,joinDuration=%s,ttl=%s)",
+		s.joinCount, s.failCount, s.createDuration, s.joinDuration, s.ttl)
+}
+
+// provisionKeyTrackInfo stores information associated with a [adjustapi.ProvisionKey]
+type provisionKeyEntry struct {
+	ttl             time.Duration
```
unmarshall, line 55: Is this correct?
In Gardener, a shoot cluster can have one or more Workers. Each worker can have one or more zones + a single instance type. Apart from that each worker can have its own value of `MachineCreationTimeout`.
This means if i have 3 workers, each worker has 1 zone + instance-type-A then that will create 3 MCDs. Now each MCD can have its own TTL which can be different. 

So for a single `provisionKeyEntry` why would you then have a single TTL? In fact it is wrong to store TTL against a set of deployment names, right?

unmarshall (changes requested):
Another set of comments. I will take more time to go over this PR.

controllers/adjuster/state.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (t machineTrackInfo) String() string {
+	return fmt.Sprintf("(provisionKey=%s,namespacedName=%s,joined=%t,failed=%t,ttl=%s)",
+		t.provisionKey, t.namespacedName, t.joined, t.failed, t.ttl)
+}
+
+// machineDeploymentStats stores aggregated observations for a MachineDeployment within a ttl.
+type machineDeploymentStat struct {
+	ttl            time.Duration
```
unmarshall: instead of TTL you could simplify it to say as long as MCD is not marked for deletion or has at least one tracked machine it will not be removed, else it will be removed.

unmarshall: instead of TTL you could simplify it to say as long as MCD is not marked for deletion or has at least one tracked machine it will not be removed, else it will be removed.

api/adjuster/types.go
elankath: It is currently set by the MCM, but not used in fallback implementation. Removing for now.

api/adjuster/types.go
elankath: The interface allows separation and non-export of implementation. The idea was to add some more methods regarding pattern of common DWD controllers. Allowing controller implementations structs to be non-exported. But this requires changes in `prober` and `weeder` too, too much work, removing it for now.  And making adjuster `Controller` struct exported.

controllers/adjuster/adjuster.go
elankath: That isn't conservatively safe. The safest conservative `ttl` is to allow objects to persist till next step.

controllers/adjuster/controller.go
elankath: No, this is not put on the `Machine` object. I can raise an MCM issue for this, but it involves some refactoring work in the MCM. Remember  instance type is technically considered a `MachineClass.ProviderSpec` which is a `runtime.RawExtension`.

controllers/adjuster/controller.go
elankath: Open to be better name. Possibly change to `instanceCreateDuration` ?  Then I will need to also rename `joinDuration` to `machineJoinDuration`. It is primarily meant for tracking and to keep things consistent as we track durations and not individual times.

controllers/adjuster/controller.go
elankath: Yes, this is known. I believe this is an OK compromise.  If you have a better suggestion, please suggest the same.

controllers/adjuster/state.go
elankath: This was intentional. If the Machine create was missed, the adjuster doesn't bother tracking the same Machine's joins. Because Machine can technically join a cluster several times.  But perhaps this can be simplified - will discuss this with you.

controllers/adjuster/state.go
elankath: Renaming to `maxInstanceCreateDuration` and `maxNodeJoinDuration` respectively.

controllers/adjuster/state.go
elankath: Renaming to `maxNodeJoinDuration`.

controllers/adjuster/helper.go
elankath: (setup meeting to discuss this)

"reconcileMachineFail will not trigger since join even resets the fail count."
Yes, this is intended. Adjuster needs to be conservative in increasing the `effective-machine-creation-timeout`. If even a single Machine successfully joins for an MCD, it should not increase the timeout for all MCD's as it has been proven that existing creation timeout can work.

controllers/adjuster/state.go
elankath: By worker, I am assuming you mean worker pool ? Sorry, Didn't understand the bit about why single TTL is bad We need a ttl as we don't want to keep building this map everytime (too much work), but we don't want to track the information forever. So the `ttl` is effectively refreshed for every call from `trackMachineCreate`, `trackMachineJoin` and `trackMachineFail`.  So the deployment names are only remember IF they are being actively used.

controllers/adjuster/state.go
elankath: Changed to "The stat is associated with a ttl that is refreshed whenever a machine belonging to the MachineDeployment is tracked."

controllers/adjuster/state.go
elankath: Hmm..on second though, it might make sense to ONLY track machine joins and not bother about machine creates, since it is effectively irrelevant to the `effective-creation-timeout`. It was only being done for tracking/informational purpose. And since lookup of MachineClass is cheap as it is in the informer cache, I think I will just drop `reconcileMachineCreate` completely.

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	// AnnotationKeyEffectiveCreationTimeout is the annotation key set on the MachineDeployment that indicates
+	// the effective creation timeout for all Machine's belonging to this MachineDeployment. If specified, the value for this
+	// annotation takes precedence over the MachineDeployment.Spec.Template.Spec.MachineCreationTimeout.
+	AnnotationKeyEffectiveCreationTimeout = "node.machine.sapcloud.io/effective-creation-timeout"
+	// AnnotationKeyEffectiveCreationTimeoutLastAdjustedAt is the annotation key set on the MachineDeployment that indicates
+	// the time that the effective creation timeout was last adjusted for this MachineDeployment.
+	AnnotationKeyEffectiveCreationTimeoutLastAdjustedAt = "node.machine.sapcloud.io/effective-creation-timeout-last-adjusted-at"
+	//DefaultCreationTimeoutGrowthFactor is the default value for growth of the effective-creation-timeout.
```
takoverflow, line 22:
```suggestion
	// DefaultCreationTimeoutGrowthFactor is the default value of the multiplier used for the growth of effective-creation-timeout.
```

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	DefaultCreationTimeoutGrowthFactor = 2.0
+	//DefaultCreationTimeoutMax is the max limit beyond upto which the machine-creation-timeout can be adjusted
```
takoverflow, line 24:
```suggestion
	// DefaultCreationTimeoutMax is the maximum limit upto which the machine-creation-timeout can be adjusted
```

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	DefaultCreationTimeoutMax = 90 * time.Minute
+	// DefaultMachineFailureThresholdFraction is the default value of the threshold fraction for Failed Machines corresponding
+	// to a [adjustapi.ProvisionKey]. Once breached, the adjuster will revise the effective machine-creation-timeout upwards
+	// by the [CreationTimeoutGrowthFactor].
+	DefaultMachineFailureThresholdFraction = 0.2
+	// DefaultMachineFailureThresholdMin default value of the minimum number for Failed Machines corresponding
```
takoverflow, line 30:
```suggestion
	// DefaultMachineFailureThresholdMin is the default value of the minimum number for Failed Machines corresponding
```

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	// to an [adjustapi.ProvisionKey]. Only once this is surpassed, will the failure fraction be computed for comparing
+	// against MachineFailureThresholdFraction.
+	DefaultMachineFailureThresholdMin = 2
```
takoverflow, line 33: ~~Just for my understanding, the min here needs to be surpassed as per the docstring, i.e. if there are 2 failing machines the adjuster still doesn't kick in. Only once it hits 3 machines for the same `ProvisionKey` and the `FailureThresholdFraction` is breached i.e. `>20%` of failing machines does the adjuster start its computation of new effective-creation-timeout?~~

Nvm, checked the definition of `HasBreached`
```go
// HasBreached confirms whether the failCount > thresholdMin and the fraction failCount/failCount+jointCount has breached
// the given thresholdFraction.
func (s machineProvisionKeyStat) HasBreached(thresholdMin uint32, thresholdFraction float64) bool {
	if s.failCount < thresholdMin {
		return false
	}
	total := s.failCount + s.joinCount
	if total == 0 {
		return false
	}
	return float64(s.failCount)/float64(total) >= thresholdFraction
}
```

The docstrings here and for HasBreached are incorrect, the actual function returns true when `failCount = thresholdMin` assuming the fraction also is `>=` the threshold. Please fix the docstrings to reflect the same.

> Only once this is surpassed

This needs to be corrected as well.

api/adjuster/types.go
takoverflow: Alternatively it could also be called `DefaultMachineCreationTimeout`, but I agree that the `Default` prefix should be there.

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	// StandardCreationTimeout is the standard machine creation timeout in gardener clusters if not overridden.
+	// See https://github.com/gardener/gardener/blob/e140ccc402b8732499cb804190fc4fe1ce82c078/example/90-shoot.yaml#L142
+	StandardCreationTimeout = 20 * time.Minute
+)
+
+// Config provides typed access to adjuster configuration. Corresponds to the config map used to configure the adjuster controller
```
takoverflow, line 39:
```suggestion
// Config provides typed access to adjuster configuration. Corresponds to the ConfigMap used to configure the adjuster controller
```
Nit

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+type Config struct {
+	// CreationTimeoutGrowthFactor is the growth factor used by adjuster for increasing the effective machine-creation-timeout
+	// If unspecified, the default value is [DefaultCreationTimeoutGrowthFactor].
+	CreationTimeoutGrowthFactor *float64 `json:"creationTimeoutGrowthFactor"`
+	// CreationTimeoutMax is the maximum effective machine-creation-timeout set by the adjuster on MachineDeployment objects.
```
takoverflow, line 44:
```suggestion
	// CreationTimeoutMax is the maximum effective machine-creation-timeout that can be set by the adjuster on MachineDeployment objects.
```
Nit

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	// If unspecified, the default value is [DefaultCreationTimeoutMax]
+	CreationTimeoutMax *metav1.Duration `json:"creationTimeoutMax"`
+	// MachineFailureThresholdFraction is the fraction  for Failed Machines corresponding to a [adjustapi.ProvisionKey].
```
takoverflow, line 47:
```suggestion
	// MachineFailureThresholdFraction is the fraction of Failed Machines corresponding to a [adjustapi.ProvisionKey].
```

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	// Once breached, the adjuster will revise the effective machine-creation-timeout upwards by the
+	// [CreationTimeoutGrowthFactor].
+	// If unspecified, the default value is [DefaultFailureThreshold]
+	MachineFailureThresholdFraction *float64 `json:"machineFailureThresholdFraction"`
+	// MachineFailureThresholdMin indicates the minimum Machine failures corresponding to an [adjustapi.ProvisionKey]
+	// only after which the fraction threshold is considered.
```
takoverflow, line 53:
```suggestion
	// after which the failure fraction threshold is considered for creation timeout adjustment.
```

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	MachineFailureThresholdMin *uint32 `json:"machineFailureThresholdMin"`
+}
+
+// MachineProvisionKey identifies a group of machines with the same instance
+// type and availability zone for which provisioning statistics are tracked.
+//
+// Machines with the same instance type and availability zone share the same
```
takoverflow, line 60:
```suggestion
// These groups of machines share the same
```
Nit: No need to repeat the characteristics here, it's already mentioned above.

api/adjuster/types.go
```diff
@@ -0,0 +1,70 @@
...
+	// to an [adjustapi.ProvisionKey]. Only once this is surpassed, will the failure fraction be computed for comparing
+	// against MachineFailureThresholdFraction.
+	DefaultMachineFailureThresholdMin = 2
+	// StandardCreationTimeout is the standard machine creation timeout in gardener clusters if not overridden.
+	// See https://github.com/gardener/gardener/blob/e140ccc402b8732499cb804190fc4fe1ce82c078/example/90-shoot.yaml#L142
+	StandardCreationTimeout = 20 * time.Minute
+)
+
+// Config provides typed access to adjuster configuration. Corresponds to the config map used to configure the adjuster controller
+type Config struct {
```
takoverflow, line 40: All the `Config` fields' json struct tags are missing `omitempty`, since these are all optional fields I'm guessing that should be present.

cmd/probercmd.go
```diff
@@ -24,8 +25,9 @@ import (
 )
 
 const (
-	proberLeaderElectionID = "dwd-prober-leader-election"
-	weederLeaderElectionID = "dwd-weeder-leader-election"
+	proberLeaderElectionID   = "dwd-prober-leader-election"
+	weederLeaderElectionID   = "dwd-weeder-leader-election"
+	adjusterLeaderElectionID = "dwd-adjuster-leader-election"
```
takoverflow, line 30: `weededLeaderElectionID` and `adjusterLeaderElectionID` should be in their corresponding `cmd.go` files, not in `probercmd.go`

api/prober/.import-restrictions
```diff
@@ -6,4 +6,4 @@ rules:
   - selectorRegexp: github[.]com/gardener/dependency-watchdog
     allowedPrefixes:
     # should be self-contained and must not import any other dependency watchdog packages
-      - github.com/gardener/dependency-watchdog/api/prober
\ No newline at end of file
+      - github.com/gardener/dependency-watchdog/api/adjuster
```
takoverflow, line 9: Shouldn't this be prober, was there a reason why this was changed?

cmd/adjustercmd.go
```diff
@@ -0,0 +1,119 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
+)
+
+var (
+	// AdjusterCmd stores info about using the adjuster command
+	AdjusterCmd = &Command{
+		Name:      "adjuster",
+		UsageLine: "",
+		ShortDesc: "Checks Machine metrics/status and adjusts machine-creation-timeout on MachineDeployment's",
```
takoverflow, line 29:
```suggestion
		ShortDesc: "Checks Machine metrics/status and adjusts machine-creation-timeout on MachineDeployments",
```
Nit

cmd/adjustercmd.go
takoverflow: Also, the current default TimeoutMax value is `1.5h` not `2h`.

cmd/adjustercmd.go
```diff
@@ -0,0 +1,119 @@
...
+		Run:      startAdjusterController,
+	}
+	adjusterOpts = adjusterOptions{}
+)
+
+type adjusterOptions struct {
+	SharedOpts
+}
+
+func init() {
```
takoverflow, line 70: Is this `init` needed? It's adding to the scheme defined in `probercmd` and there we're already registering these same types.

Additionally, should the scheme construction and registration be done outside of `probercmd`, since it's not exclusive to the prober, WDYT?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,211 @@
...
+	FillConfigDefaults(config)
+	err = ValidateConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+// ValidateConfig validates the given adjuster Configuration.
+func ValidateConfig(c *adjustapi.Config) error {
```
takoverflow, line 40: Just a minor nit, but it's easier to read a file when the functions are ordered by the sequence in which they are invoked. So would it be possible to move this below `fillConfigDefaults`?

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,211 @@
...
+	if c.CreationTimeoutMax != nil {
+		v.MustNotBeZeroDuration("CreationTimeoutMax", *c.CreationTimeoutMax)
+	}
+	if c.CreationTimeoutGrowthFactor != nil {
+		v.MustBeGreater("CreationTimeoutGrowthFactor", *c.CreationTimeoutGrowthFactor, 1.0)
+	}
+	if v.Error != nil {
+		return v.Error
+	}
+	return nil
```
takoverflow, line 51:
```suggestion
	return v.Error
```

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	} else {
+		config = &adjustapi.Config{}
+	}
+	FillConfigDefaults(config)
+	err = ValidateConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
```
takoverflow, line 36:
```suggestion
	return
```
Since named return args are being used

controllers/adjuster/helper.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+// LoadConfig reads the adjuster configuration from a file if specified, unmarshalls into instance of
+// [adjustapi.Config], fills default values for missing config fields using [FillConfigDefaults] and returns
+// populated object or an error.
+func LoadConfig(configPath string) (config *adjustapi.Config, err error) {
+	if configPath != "" {
+		config, err = util.ReadAndUnmarshall[adjustapi.Config](configPath)
+		if err != nil {
+			return nil, err
```
takoverflow, line 26:
```suggestion
			return
```

takoverflow (commented):
Thanks for the PR!

Still going through the entire thing, this is just the first set of comments, PTAL

Mostly consists of nits and docs cleanup.

controllers/adjuster/adjuster.go
```diff
@@ -0,0 +1,340 @@
...
+			deploymentStats:   cache.NewExpiring(),
+			machineTrackInfos: cache.NewExpiring(),
+			provisionKeyStats: cache.NewExpiring(),
+		},
+	}
+}
+
+// Reconcile listens to filtered Update events for [machinev1alpha1.Machine] resources and adjusts effective `machine-creation-timeout`s
+// on [machinev1alpha1.MachineDeployment]'s.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
takoverflow: I see that the `Prober` and `Weeder` controllers have `kubebuilder rbac` tags above them specifying the kinds of resources those controllers would need access to. Shouldn't there be something similar for the the `Adjuster` if we're going to be listing/fetching/updating machine and MCD objects?

